PR #9, opened 2024-10-28
Chore/add unified linting

robert-northmind:
This PR is an initial bigger clean-up/organizing task. This is roughly what happened 

- Adds shared analyzer file for all flutter packages
- Updates all source files to conform to new linting rules
- Adds VSCode config file for easy running example app from VS code
- Fix crash when starting example app
- Fix broken unit tests

### DIFF
--- a/packages/rum_sdk/example/lib/main.dart
+++ b/packages/rum_sdk/example/lib/main.dart
@@ -73,7 +73,7 @@ class HomePage extends StatefulWidget {
   const HomePage({Key? key}) : super(key: key);
 
   @override
-  _HomePageState createState() => _HomePageState();
+  State<HomePage> createState() => _HomePageState();
 }
 
 class _HomePageState extends State<HomePage> {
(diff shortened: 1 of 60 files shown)


packages/rum_sdk/example/lib/main.dart
```diff
@@ -4,25 +4,27 @@ import 'dart:io';
...
 import 'package:http/http.dart' as http;
 
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
   HttpOverrides.global = RumHttpOverrides(HttpOverrides.current);
-  await dotenv.load(fileName: ".env");
-  RumFlutter()
-      .transports
-      .add(OfflineTransport(maxCacheDuration: const Duration(days: 3), collectorUrl: dotenv.env['FARO_COLLECTOR_URL'] ?? ''));
+
```
robert-northmind, line 12: Using the `dotEnv` package is not that safe since it puts the raw .env file in the build app. To lead with a better example, I think it is better to showcase a usage of the flutter provided json config instead .

packages/rum_sdk/example/lib/main.dart
```diff
@@ -73,7 +75,7 @@ class HomePage extends StatefulWidget {
   const HomePage({Key? key}) : super(key: key);
 
   @override
-  _HomePageState createState() => _HomePageState();
+  State<HomePage> createState() => _HomePageState();
```
robert-northmind, line 78: Since the `_HomePageState` is private, it is a better approach to not expose it and define this as `State<HomePage>`

Issue #5, opened 2024-10-28
Investigate how we could add tracing support


robert-northmind:
The Faro protocol supports tracing also. But the framework does not yet support it.
We could initially use Otel-dart to get that support working. It would be really benefitial to have this up an running as it connects backend and frontend debugging.

robert-northmind:
Resolved by PR #17

PR #17, opened 2024-12-13
Add tracing support using otel

robert-northmind:
This PR adds support for traces.

Traces are collected under the hood using the [opentelemetry](https://pub.dev/packages/opentelemetry) package.
A custom Faro-exporter extracts the traces out from Otel and sends them over the existing faro protocol.

The external opentelemetry dependency is hidden and the public Faro-Trace-API exposes its own types.
This makes it safe in the future to change the underlying implementation without a need to change the API.

This is the first version of tracing support. It supports:
- auto instrumenting all http requests
- manually passing a parent span when creating a new span
- or setting a span to "active", and all spans created later will inherit that span as a parent.

Later we will probably add zone-support for inferring parent spans.

This PR solves #5

### DIFF
--- a/packages/rum_sdk/lib/src/tracing/dart_otel_tracer_resources_factory.dart
+++ b/packages/rum_sdk/lib/src/tracing/dart_otel_tracer_resources_factory.dart
@@ -1,0 +1,47 @@
+import 'package:opentelemetry/api.dart' as otel_api;
+import 'package:opentelemetry/sdk.dart' as otel_sdk;
+import 'package:rum_sdk/rum_flutter.dart';
+
+class DartOtelTracerResourcesFactory {
+  otel_sdk.Resource getTracerResource() {
+    final rumFlutter = RumFlutter();
+    const unknownString = 'unknown';
+
+    return otel_sdk.Resource(
+      [
+        // App info
+        otel_api.Attribute.fromString(
+          otel_api.ResourceAttributes.serviceName,
+          rumFlutter.meta.app?.name ?? unknownString,
+        ),
+        otel_api.Attribute.fromString(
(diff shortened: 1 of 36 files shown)


packages/rum_sdk/lib/src/tracing/dart_otel_tracer_resources_factory.dart
```diff
@@ -0,0 +1,47 @@
...
+          otel_api.ResourceAttributes.serviceVersion,
+          rumFlutter.meta.app?.version ?? unknownString,
+        ),
+
+        // Faro info
+        otel_api.Attribute.fromString('faro.sdk.name', 'faro-flutter-sdk'),
+        otel_api.Attribute.fromString('faro.sdk.language', 'dart'),
+        otel_api.Attribute.fromString(
+          'faro.sdk.version',
+          '1.0.0',
```
robert-northmind, line 31: This is hardcoded for now. In the future we will need to find a way to compute this.

packages/rum_sdk/lib/src/tracing/dart_otel_tracer_resources_factory.dart
```diff
@@ -0,0 +1,47 @@
...
+
+        // Otel info
+        otel_api.Attribute.fromString(
+          'telemetry.sdk.name',
+          'opentelemetry-dart',
+        ),
+        otel_api.Attribute.fromString('telemetry.sdk.language', 'dart'),
+        otel_api.Attribute.fromString(
+          'telemetry.sdk.version',
+          '0.18.6',
```
robert-northmind, line 42: The same here.

PR #29, opened 2025-01-23
Remove already reported crash reports on android

robert-northmind:
On Android, we use `activityManager.getHistoricalProcessExitReasons(...)` to get the latests crash reports.
These are stored on device and are persisted. I think it keeps the last 100 crash reports.

We were asking android for these reports on each start of our package. So we kept getting the same reports.

To fix this. I now store the handled reports in sharedPrefs. And then ignore these reports when I grab the next report.

### DIFF
--- a/packages/rum_sdk/android/src/main/java/com/example/rum_sdk/ANRTracker.java
+++ b/packages/rum_sdk/android/src/main/java/com/example/rum_sdk/ANRTracker.java
@@ -2,21 +2,15 @@ package com.example.rum_sdk;
 
 import android.os.Handler;
 import android.os.Looper;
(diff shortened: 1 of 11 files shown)


packages/rum_sdk/android/src/main/java/com/example/rum_sdk/ANRTracker.java
```diff
@@ -2,22 +2,16 @@
 
 import android.os.Handler;
 import android.os.Looper;
-import android.provider.Settings;
```
robert-northmind, line 5: not needed imports

imtoori (approved):

PR #33, opened 2025-02-18
fix: add custom user updated event when send user meta info

yduartep:
add custom user updated event when send user meta info

### DIFF
--- a/packages/rum_sdk/lib/rum_flutter.dart
+++ b/packages/rum_sdk/lib/rum_flutter.dart
@@ -161,7 +161,7 @@ class RumFlutter {
         Meta.fromJson({..._instance.meta.toJson(), 'user': userMeta.toJson()});
     _instance._batchTransport?.updatePayloadMeta(_instance.meta);
 
-    RumFlutter().pushEvent('faro_internal_user_updated')
+    RumFlutter().pushEvent('faro_internal_user_updated');
   }
 
   void setViewMeta({String? name}) {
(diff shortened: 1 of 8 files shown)


packages/rum_sdk/lib/rum_flutter.dart
```diff
@@ -160,6 +160,14 @@ class RumFlutter {
     _instance.meta =
         Meta.fromJson({..._instance.meta.toJson(), 'user': userMeta.toJson()});
     _instance._batchTransport?.updatePayloadMeta(_instance.meta);
+
+    if (userId != null || userName != null || userEmail != null) {
+      RumFlutter().pushEvent('user_updated', attributes: {
```
robert-northmind, line 165: 2 thoughts:

1: maybe we should prefix our internal Faro-events with something `faro_internal_{the event name}`? 
to avoid that a user of the SDK accidentally creates the same event..?

2: Should we send the 
`RumFlutter().pushEvent('user_updated' ..)`-event even if the userId is empty maybe?
I mean, it could be that the user logs out? In which case we want to record that the user is now "empty"

what do you think?

packages/rum_sdk/lib/rum_flutter.dart
```diff
@@ -160,6 +160,14 @@ class RumFlutter {
...
+        'userId': userId ?? '',
+        'userName': userName ?? '',
+        'userEmail': userEmail ?? ''
```
robert-northmind, line 168: I think also, we don't really have to send the actual data here.
Since the data will be in the meta data already attached to this event?

yduartep: yeah It is true

robert-northmind (dismissed):

robert-northmind (approved):

Issue #12, opened 2024-10-29
Offline_transport never clears cached logs


robert-northmind:
When the device goes offline, then all telemetry is stored to disk. 
And as soon as the device is online again, then all those telemetry is sent to the collector.
**But** there is a bug here. We never clear telemetry which has been sent.
So we keep sending the same telemetry over and over when the device is offline.

The offline cache just grows and grows.

We need to clear up the cached file as soon as we have sent the data.

Issue #11, opened 2024-10-29
Internet connectivity check in Offline_transport not reliable


robert-northmind:
The internet connectivity check in the Offline_transport is not so stable.
It checks the internet connection based on a url you pass in, which is currently the collector url. 
It checks it using `InternetAddress.lookup` but this failes with a `SocketException: Failed host lookup:`

We need to fix this. I guess here are 2 approaches:

1. Do a real request to the collectorUrl, but also provide the apiKey. or somehow ping it.
2. Change the url we use to pass to the `InternetAddress.lookup`. There are some standard ones you can use for internet connectivity checks such as: `https://one.one.one.one`, `http://www.google.com/generate_204`, `http://www.example.com`, `http://captive.apple.com`, or for china and Russia: `http://qq.com (Tencent’s portal)`, `http://yandex.ru (Yandex’s main page) `

robert-northmind:
Update after implementation + review.

**What we shipped ([#282](https://github.com/grafana/faro-flutter-sdk/pull/282)):** a targeted hardening of the DNS probe in `InternetConnectivityService` — a 5s timeout on the lookup and catching all probe errors (not just `SocketException`), keeping "probe failure ⇒ offline" so cached data is preferred over dropped data. This removes the "probe hangs → online/offline decisions stall" failure mode.

**Why this doesn't fully close the issue.** Digging in, the reliability of the connectivity *probe* is only part of the story. The bigger risks are in how send outcomes are handled, and a "perfect probe" can't fix them:

1. **Online sends are fire-and-forget.** `FaroTransport.send` logs and swallows errors — it neither retries nor re-caches on failure. If the probe says "online" but the actual POST fails (transient 5xx, drop, TLS), that telemetry is lost regardless of how good the probe is.
2. **Offline flush can delete data on failed sends.** `OfflineTransport._sendCachedData` always returns `true` (because `FaroTransport.send` never throws), so `_readFromFile` clears flushed lines from disk even if the underlying POST failed — a silent data-loss path.
3. **The VPN / split-horizon trap.** Probing a generic host (`one.one.one.one`) can report "offline" when the collector is actually reachable over a private network, and vice-versa. Prediction-based gating is inherently guessy.

**Proposed direction for the follow-up: outcome-driven delivery.** Instead of predicting connectivity and then blindly sending, make delivery react to real send outcomes:

- `send` returns a typed result (success / transient failure / permanent failure).
[...]

Issue #10, opened 2024-10-29
Handle Write and Read to file better


robert-northmind:
In the Offline_Transport package we are not handling writing and reading to file correctly.
We have multiple writes to file before the previous one completes.
We need to sync access to the write and read logic.
This logic is inside the `offline_transport.dart` file

Issue #4, opened 2024-10-28
Add linting and basic CI/CD workflow


robert-northmind:
To ensure the we try to follow the same coding standards in the project, we should add some linting rules, 
and also add some GitHub action CI/CD verification steps which can run on each PR

robert-northmind:
This issue was fixed by this PR https://github.com/grafana/faro-flutter-sdk/pull/9

PR #45, opened 2025-04-29
feat: Initial Flutter Web support & Tracing Exploration

pingebat:
Hey everyone,

This PR represents an initial effort to extend the `faro-flutter-sdk` to support the Flutter Web platform, alongside the existing native platforms (Android, iOS, etc.). The original SDK focused on native capabilities, and this work aims to make Faro usable in web environments as well.

**Motivation:**

The goal is to allow developers using Faro to instrument their Flutter applications regardless of the target platform, providing a unified observability solution across web and native.

**Summary of Changes:**

To achieve cross-platform compatibility and start adding web features, the main strategies I used were:

1.  **Platform Guarding:** Used `kIsWeb` (from `flutter/foundation.dart`) and a adapted `PlatformInfoProvider` to conditionally execute code paths specific to native platforms. I first focused on getting the SDK to compile and run on the web by guarding native-only calls (like those using `_nativeChannel` or `dart:io`).
2.  **Conditional Imports:** Leveraged Dart's conditional import mechanism along with **stub files** to prevent the compiler from attempting to load platform-specific libraries on incompatible targets. Examples include:
    *   Preventing `dart:io` imports (`InternetAddress`, `Directory`) on the web.
    *   Preventing web-specific imports (`package:http/browser_client.dart`, `package:opentelemetry/web_sdk.dart`) on native platforms.
3.  **Web-Specific Implementations:**
    *   Introduced `FaroWebHttpClient` (conditionally imported) to eventually handle automatic trace context propagation for HTTP requests made on the web.
    *   Configured the OpenTelemetry `TracerProvider` to use the `WebTracerProvider` on web and the standard `TracerProviderBase` on native.
    
With just steps 1 and 2, a lot of features available on native platforms also work on the web, including logs for route changes and custom events.
 
**Web HTTP Tracing Implementation Focus:**

A key focus of this initial web work was exploring how to implement automatic HTTP request tracing, and trying to keep the developer experience similar to the native `HttpOverrides` approach. Since `dart:io` (and thus `HttpOverrides`) isn't available on the web, and `package:http` typically uses `BrowserClient` (XHR), I investigated using `http.runWithClient`. This method allows scoping a custom `http.Client` (like `FaroWebHttpClient`) for code executed within its closure, seemed like a promising way to implicitly add tracing headers, similar to how `HttpOverrides` works for native.

The current implementation makes `FaroWebHttpClient` available via `Faro.instance.createHttpClient()` and uses `http.runWithClient` inside the `Faro.runApp` method (on web only) to wrap the user's application startup. This means that `Faro().runApp(...)` interface stays the same.

**Zone Mismatch Problem**

While the code now *compiles* for both platforms, a significant challenge remains when running on **web** with the current structure, specifically the **"Zone mismatch" error**.

*   **The Goal:** Maintain user flexibility by allowing `WidgetsFlutterBinding.ensureInitialized()` to be called early in the user's `main.dart`. This lets users perform setup tasks that might require bindings (like using platform channels, loading assets) before Faro initialization or `runApp`.
*   **The Conflict:**
    1.  The user calls `WidgetsFlutterBinding.ensureInitialized()` in their `main` function (this happens in the initial "root" [zone](https://api.dart.dev/stable/3.7.2/dart-async/Zone-class.html), **Zone A**).
    2.  The current `Faro.runApp` method, when running on web, uses `http.runWithClient(...)` to wrap the subsequent initialization and the final Flutter `runApp` call. This `runWithClient` creates a **new Zone (Zone B)** to scope the `FaroWebHttpClient`.
    3.  Inside Zone B, `Faro.runApp` calls Flutter's `runApp(child)`.
    4.  Flutter checks if the current zone (`Zone.current`, which is Zone B) is the same as the zone where the bindings were first initialized (Zone A). Since `Zone B != Zone A`, the assertion fails.
*   **The Error:** This results in the following error on web when `ensureInitialized` is called in `main`:
    ```
     ═════════════════════════════════Exception caught by Flutter framework ═════════════════════════════════
    The following assertion was thrown during runApp:
    Zone mismatch.
    The Flutter bindings were initialized in a different zone than is now being used. This will likely cause confusion and bugs as any zone-specific configuration will inconsistently use the configuration of the original binding initialization zone or this zone based on hard-to-predict factors such as which zone was active when a particular callback was set.
    It is important to use the same zone when calling `ensureInitialized` on the binding as when calling `runApp` later.
    To make this warning fatal, set BindingBase.debugZoneErrorsAreFatal to true before the bindings are initialized (i.e. as the first statement in `void main() { }`).

    When the exception was thrown, this was the stack:
    #0      BindingBase.debugCheckZone.<anonymous closure> (package:flutter/src/foundation/binding.dart:505:29)
    #1      BindingBase.debugCheckZone (package:flutter/src/foundation/binding.dart:510:6)
    #2      _runWidget (package:flutter/src/widgets/binding.dart:1494:18)
    #3      runApp (package:flutter/src/widgets/binding.dart:1430:3)
    #4      Faro.runApp.<anonymous closure>.<anonymous closure> (package:faro/faro.dart:160:21)  <- Inside runWithClient's callback
    #5      Faro.runApp.<anonymous closure> (package:faro/faro.dart:156:9)
    #10     runWithClient (package:http/src/client.dart:230:5)                  <- The zone creator
    #11     Faro.runApp (package:faro/faro.dart:154:5)
    #12     main (package:faro_example/main.dart:23:3)                          <- Where Faro.runApp was called
    ```
*   **Further Reading:** [Flutter Docs on Zone Mismatch](https://docs.flutter.dev/release/breaking-changes/zone-errors) explains this constraint.

I also experimented with bundling and running the official Faro Web JS SDK alongside the Flutter app, which *did* work for capturing requests, but it introduced significant build process complexity and resulted in two separate Faro sessionIds (one from JS, one from Dart), which felt suboptimal.

**Potential Alternatives:**

If the approach discussed previously cannot be successfully implemented, then based on my current understanding and experimentation, resolving the Zone Mismatch for web means choosing one of the outlined approaches below, both involving trade-offs compared to the ideal scenario. There may be other solutions which did not occur to me. 

1.  **Approach 1: Separate `initialize` and `runApp` (Prioritizes User Flexibility)**
    *   **How:** Faro would provide `Faro.instance.initialize(...)`. The user would call `ensureInitialized()` -> user setup -> `Faro.initialize()` -> standard `runApp()` in their `main`. `Faro.initialize` would *not* use `runWithClient`.
    *   **Pro:** Maximum user flexibility for setup code after `ensureInitialized`. No Zone Mismatch.
    *   **Con:** Web HTTP tracing would *not* be implicit. Users *must* explicitly use `Faro.instance.createHttpClient()` for traced web requests. The API diverges slightly from a single `runApp` wrapper. (See example: [dart-lang/http flutter\_http\_example](https://github.com/dart-lang/http/tree/master/pkgs/flutter_http_example))

2.  **Approach 2: Combined `Faro.runApp` (Prioritizes API Simplicity & Web Tracing)**
    *   **How:** Keep the single `Faro.instance.runApp(...)` API. Internally, it *conditionally* uses `runWithClient` on web, and critically, it must also handle the `ensureInitialized()` call *inside* that zone *before* calling the framework's `runApp`.
    *   **Pro:** Simplest API for the user (`Faro.runApp`). Implicit web HTTP tracing works via `http.get/post` etc.
    *   **Con:** Significantly restricts users. They *cannot* reliably run setup code that depends on Flutter bindings *before* calling `Faro.runApp`.

**Seeking Feedback:**

This PR adds web support and begins the web tracing implementation. But it **currently triggers the Zone Mismatch error** if `ensureInitialized` is called early in main. 

I've reached the limit of my knowledge on how to best resolve this Zone/initialization conflict while maintaining both user flexibility and a simple, consistent API like the original native-focused SDK had. I hope someone else can chime in to help figure out how to implement the current approach, as the two outlined alternatives have significant drawbacks.

I would greatly appreciate feedback on:

*   The overall strategy for web compatibility (conditional imports, stubs, platform checks).
*   **Crucially:** Guidance on resolving the **Zone Mismatch** issue. Is there a way to reconcile early `ensureInitialized` by the user with the web's `runWithClient` scoping that I've missed, or must we choose between Approach 1 and Approach 2? If so, which trade-off is preferred for this library?
*   Is there maybe a completely different way to intercept http requests on the web? 
*   The implementation details of `FaroWebHttpClient` and the web OTel provider setup.
*   Any other potential improvements or considerations for Flutter Web support.

I'm happy to learn and contribute, but feel like I need input from more experienced Flutter developers on the best path forward regarding the Zone issue.

Thank you for taking the time to review this! 

(This is my first time contributing to an open source project, so please let me know if you have any feedback or suggestions.)


robert-northmind:
Hey @pingebat, huge thanks for your contribution. We really appreciate this 🤩 
I'm the core maintainer of this repo, and I am heading off for a 1 week vacation starting today. 
So unfortunately I will not have time to review this now. But I will go over this as soon as I am back again.

again, big thanks for your contribution. And what a great PR description 👌 🤩

lib/src/offline_transport/internet_connectivity_service.dart
```diff
@@ -63,11 +65,13 @@ class InternetConnectivityService {
...
   Future<bool> _isConnectedToInternet() async {
     try {
-      final result = await InternetAddress.lookup(
-        _internetConnectionCheckerUrl,
-      );
-      return result.isNotEmpty && result[0].rawAddress.isNotEmpty;
-    } on SocketException catch (_) {
+      if (kIsWeb) {
+        final connectivityResult = await Connectivity().checkConnectivity();
+        return !connectivityResult.contains(ConnectivityResult.none);
```
robert-northmind, line 70: Is this checking similar logic as the mobile platform?
Like the mobile platform is supposed here to check if it actually has a connection to the internet. not only if it has a connection to a wifi etc..

lib/src/offline_transport/internet_connectivity_service.dart
```diff
@@ -63,11 +65,13 @@ class InternetConnectivityService {
...
+      } else {
+        return await performNativeInternetCheck(_internetConnectionCheckerUrl);
```
robert-northmind, line 72: Instead of here doing this `if (kIsWeb)`-check, could we try to use some polymorphism? and have some abstract class maybe which defines a shared interface? and then we could have a factory inject the correct concrete class?

robert-northmind (commented):
Hey @pingebat, sorry for the delay. It has been a busy week.
But I started reviewing you PR now and left some initial comments.
I didn't have time yet to go through it all. But I will continue

lib/faro.dart
```diff
@@ -116,18 +124,20 @@ class Faro {
...
+    if (!_platformInfoProvider.isWeb) {
+      if (Platform.isAndroid || Platform.isIOS) {
+        NativeIntegration.instance.init(
+            memusage: optionsConfiguration.memoryUsageVitals,
+            cpuusage: optionsConfiguration.cpuUsageVitals,
+            anr: optionsConfiguration.anrTracking,
+            refreshrate: optionsConfiguration.refreshRateVitals,
+            setSendUsageInterval: optionsConfiguration.fetchVitalsInterval);
+      }
+      await _instance.pushEvent('session_start');
```
robert-northmind, line 136: Why are we not sending the `session_start` event for web?

lib/faro.dart
```diff
@@ -116,18 +124,20 @@ class Faro {
...
     }
-    if (Platform.isAndroid || Platform.isIOS) {
-      NativeIntegration.instance.init(
-          memusage: optionsConfiguration.memoryUsageVitals,
-          cpuusage: optionsConfiguration.cpuUsageVitals,
-          anr: optionsConfiguration.anrTracking,
-          refreshrate: optionsConfiguration.refreshRateVitals,
-          setSendUsageInterval: optionsConfiguration.fetchVitalsInterval);
+    if (!_platformInfoProvider.isWeb) {
+      if (Platform.isAndroid || Platform.isIOS) {
```
robert-northmind, line 128: If we now have a `_platformInfoProvider`, why not use that to check isAndroid or isIOS?

lib/faro.dart
```diff
@@ -137,7 +147,14 @@ class Faro {
     OnErrorIntegration().call();
     FlutterErrorIntegration().call();
     await init(optionsConfiguration: optionsConfiguration);
-    await appRunner!();
+    if (kIsWeb){
```
robert-northmind, line 150: We have the `_platformInfoProvider.isWeb`, shouldn't we use that instead here maybe?

lib/faro.dart
```diff
@@ -116,18 +124,20 @@ class Faro {
...
-          refreshrate: optionsConfiguration.refreshRateVitals,
-          setSendUsageInterval: optionsConfiguration.fetchVitalsInterval);
+    if (!_platformInfoProvider.isWeb) {
+      if (Platform.isAndroid || Platform.isIOS) {
+        NativeIntegration.instance.init(
+            memusage: optionsConfiguration.memoryUsageVitals,
+            cpuusage: optionsConfiguration.cpuUsageVitals,
+            anr: optionsConfiguration.anrTracking,
+            refreshrate: optionsConfiguration.refreshRateVitals,
+            setSendUsageInterval: optionsConfiguration.fetchVitalsInterval);
```
robert-northmind, line 134: Are some of these metrics still possible to get from web? Like memusage and cpu? Or some other metrics important for web maybe?
Maybe for a future PR, we could think about adding some Web-NativeIntegration? 🤔

robert-northmind (commented):
I left a few more comments. And I had another thought.

First of all, huge thanks for this great contribution 🤩 
But what do you think of doing it in smaller pieces?

I am not that familiar with Flutter-web and how to use HTTP there.

What do you think about, for this initial version of adding web support, we don't add http-tracing. just the logs, event,metrics etc.

And then later, we can add http-tracing. That way, we have more time to experiment and think through that solution.

For native mobile, we also didn't have training initially, this only got added at a later stage. 

This would make this PR smaller and easier to merge. 
And then we can have a new PR specifically targeting the web tracing. 

E.g. Maybe there is some way to have some shared logic in the span-creation between the native mobile and web..

Just a thought

lib/faro.dart
pingebat: Thanks for your notes! I will try to incorporate them this week.

I think what you are proposing makes a lot of sense. 
Do you have anyone at your organization with Flutter web experience who you could reach out to? 
I have looked into how others solve this issue and they usually just provide their own http client that the user has to use. 

In regards to this specific note: I took your suggestions to leave the tracing for later so the `kIsWeb` guarded section can be removed.

robert-northmind: Hey @pingebat , cool, thanks for you effort ☺️
We don't have too many flutter experts at Grafana, and the ones we have are mainly focused on mobile.
I also had a quick look around and found similar approached to what you describe.

I know that Embrace offers support for instrumenting Dio-http clients by using a custom EmbraceInterceptor.
Maybe this is some pattern which could work. Like offering some sort of HttpClient-interceptor object.
Not sure how it would best integrate with custom http clients thought. I think we need to do some exploration and test around a bit.

But just getting logs and events etc working for web is great. And traces will be a cherry on top later 🙌

lib/src/offline_transport/internet_connectivity_service.dart
pingebat: You are right. I will create the necessary classes for this.

lib/src/offline_transport/internet_connectivity_service.dart
pingebat: I see. On web we do not have access to DNS resolution but something like this comes close I think:

``` dart
final uri = Uri.https(host);
final client = http.Client();
final request = http.Request('HEAD', uri);
final streamedResponse = await client.send(request).timeout(const Duration(seconds: 2));
```

lib/faro.dart
pingebat: So LLMs seem to say that you can collect some browser-specific things like page load, memory usage (only JS heap), FPS, and CPU (not directly through an API, only with workarounds that would impact performance itself). Most of this requires interacting with JS interop and browser APIs it appears, which I'm not knowledgeable about, so I can't really say if the suggestions of the LLMs are reasonable and these workarounds are worth the tradeoffs. Maybe a separate PR is the way to go. But happy to discuss this further.

robert-northmind: I see. good input ☺️ Then for now, let's not add any of this. 
We could in the future maybe think about adding metrics similar to how the FaroWebSDK does it (that's a javascript sdk meant for webpages). 

They collect web specific things like TTFB and FCP etc. See [here](https://grafana.com/docs/grafana-cloud/monitor-applications/frontend-observability/performance/?pg=oss-faro&plcmt=hero-btn-3#web-vitals)

that makes maybe sense.. to have flutter-web collect web specific stuff. And have flutter mobile do mobile specific metrics.

Thanks for your input and great discussions here 👌

example/lib/main.dart
```diff
@@ -11,9 +12,11 @@ void main() async {
   const faroCollectorUrl = String.fromEnvironment('FARO_COLLECTOR_URL');
   final faroApiKey = faroCollectorUrl.split('/').last;
 
-  Faro().transports.add(OfflineTransport(
-        maxCacheDuration: const Duration(days: 3),
-      ));
+  if (!kIsWeb) {
```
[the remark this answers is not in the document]
pingebat: I got an error message I did not really understand at the time so I guarded it with `kIsWeb` because it seemed to be web platform related.
Below is the stack trace:
```
DartError: MissingPluginException(No implementation found for method getApplicationDocumentsDirectory on channel plugins.flutter.io/path_provider)
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 288:3  throw_
errors.dart:288
packages/flutter/src/services/platform_channel.dart 332:7                    <fn>
platform_channel.dart:332
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 610:19          <fn>
async_patch.dart:610
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 634:23          <fn>
async_patch.dart:634
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 581:31          <fn>
async_patch.dart:581
dart-sdk/lib/async/zone.dart 1676:54                                         runUnary
zone.dart:1676
dart-sdk/lib/async/future_impl.dart 204:18                                   handleValue
future_impl.dart:204
dart-sdk/lib/async/future_impl.dart 902:44                                   handleValueCallback
future_impl.dart:902
dart-sdk/lib/async/future_impl.dart 931:13                                   _propagateToListeners
future_impl.dart:931
dart-sdk/lib/async/future_impl.dart 707:5                                    [_completeWithValue]
future_impl.dart:707
dart-sdk/lib/async/future_impl.dart 777:7                                    callback
future_impl.dart:777
dart-sdk/lib/async/schedule_microtask.dart 40:11                             _microtaskLoop
schedule_microtask.dart:40
dart-sdk/lib/async/schedule_microtask.dart 49:5                              _startMicrotaskLoop
schedule_microtask.dart:49
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 186:7           <fn>
async_patch.dart:186
```
I now stepped through it with the debugger and it originates from `offline_transport.dart` where files are being written to and read. Makes sense 😁

I guess this does mean that offline transport is not trivial to get working. 😕

lib/src/offline_transport/internet_connectivity_service.dart
pingebat: I already had the necessary classes (`NetworkReachabilityChecker` and friends) ready for this comment
> Instead of here doing this `if (kIsWeb)`-check, could we try to use some polymorphism? and have some abstract class maybe which defines a shared interface? and then we could have a factory inject the correct concrete class?

so I committed them but it might make sense to remove them again because they add (currently) an unnecessary layer of indirection. I noted this in the description of commit [085278d](https://github.com/grafana/faro-flutter-sdk/pull/45/commits/085278d2c2dc8c636f404289467d0a4b83579a95). Take a look and let me know what you think 😀

lib/src/device_info/platform_info_provider.dart
```diff
@@ -1,15 +1,19 @@
...
 
-  final operatingSystem = Platform.operatingSystem;
+  final String dartVersion = !kIsWeb ? Platform.version : 'unknown (web)';
 
-  final operatingSystemVersion = Platform.operatingSystemVersion;
+  final String operatingSystem = !kIsWeb ? Platform.operatingSystem : 'web';
+
+  final String operatingSystemVersion =
+      !kIsWeb ? Platform.operatingSystemVersion : 'unknown';
 }
```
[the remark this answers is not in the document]
robert-northmind: Nice

test/src/offline_transport/internet_connectivity_service_test.dart
```diff
@@ -19,12 +20,14 @@ void main() {
...
   late StreamController<List<ConnectivityResult>> fakeConnectivityController;
   late InternetConnectivityService service;
+  late NetworkReachabilityChecker networkReachabilityChecker;
 
   setUp(() {
     mockConnectivity = MockConnectivityChecker();
     fakeConnectivityController =
         StreamController<List<ConnectivityResult>>.broadcast();
     fakeConnectivityResults = [ConnectivityResult.none];
+    networkReachabilityChecker = NetworkReachabilityCheckerFactory().create();
```
robert-northmind, line 30: Instead of using the real `networkReachabilityChecker` here, what do you think about if we mock it, similar to how we mock the `ConnectivityChecker`?

we could do something like

```
class MockNetworkReachabilityChecker extends Mock implements NetworkReachabilityChecker {}
...
when(() => mockNetworkReachabilityChecker. isConnectedToInternet(any))
         .thenAnswer((_) async => fakeIsConnected);
```

robert-northmind (commented):
Added some minor comments

micaelcid:
Hello @robert-northmind, I noticed you closed #181. Do you have any plans to get back it? 

Otherwise, I could continue from where you stopped. Please, let me know your thoughts and concerns.

robert-northmind:
> Hello @robert-northmind, I noticed you closed #181. Do you have any plans to get back it?
> 
> Otherwise, I could continue from where you stopped. Please, let me know your thoughts and concerns.

Hi @micaelcid , We did some underlying groundwork and also included a new Dart OTel package to be used in FaroFlutterSDK. I think all of that impacted this work. So I decide to close #181  until that was done and then start out clean with this implementation. 
I am OutOfOffice for the next 2 weeks. After that I can start to look into this again.

PR #227, opened 2026-05-08
feat(tracing): add span exception handling control

Rloaded:
## Description

Add capabilities to `startSpan` for controlling how exceptions are recorded on spans:

- **`spanExceptionReporter` callback**: Optional parameter on `startSpan()` that
  replaces the default `setStatus(error)` + `recordException` catch-block behavior,
  giving users full control over span-level error recording. Exceptions are always
  rethrown — the callback controls span recording, not error propagation.

## Related Issue(s)

N/A

## Type of Change

- [x] 📝 Documentation
- [ ] 🏗️ Code refactoring
- [ ] 🧹 Chore / Housekeeping

## Checklist

### Changes by file

| File | Change |
|------|--------|
| `lib/src/tracing/span.dart` | `SpanExceptionReporter` typedef |
| `lib/src/tracing/faro_zone_span_manager.dart` | `executeWithSpan` catch block: callback branching |
| `lib/src/tracing/faro_tracer.dart` | Thread `spanExceptionReporter` parameter |
| `lib/src/faro.dart` | Thread parameter + inline API docs + usage example |
| `doc/Reference.md` | New "Custom Span Error Handling" section |
| `CHANGELOG.md` | Entry under `[Unreleased]` → `Added` |
| `test/src/tracing/faro_zone_span_manager_test.dart` | +7 tests (default path, guard path, callback path, callback safety) |
| `test/src/tracing/faro_tracer_test.dart` | +2 tests (parameter threading, null default) |

**Test results**: 600/600 pass · `flutter analyze` clean · `dart format` clean

---

> [!NOTE]
> **Medium Risk**
> Changes span exception capture behavior in the tracing hot path and introduces a new sanitization hook, which could affect what error data is recorded (or omitted) and how spans are marked on failures.
> 
> **Overview**
> Adds configurable control over how `startSpan()` handles thrown exceptions via new `SpanExceptionOptions`, supporting *selective status setting*, *selective exception recording*, and an optional `exceptionSanitizer` for PII-safe span error reporting.
> 
> Wires these options through `FaroConfig` (global default) and a new `exceptionOptions` parameter on `Faro().startSpan()` (per-span, merged field-by-field over global), and updates `FaroZoneSpanManager` to apply the effective behavior—including a safe fallback that marks the span as failed if the sanitizer itself throws.
> 
> Exports the new API, updates docs/changelog, and adds test coverage for option threading, merge semantics, and the new error-handling branches.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 75779ad54085e4e9ff318fa0d5a4132be84ce27e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/src/faro.dart
+++ b/lib/src/faro.dart
@@ -448,6 +448,13 @@ class Faro {
   ///   - [ContextScope.callback] (default): Deactivated when callback completes.
   ///   - [ContextScope.zone]: Stays active for timers/streams in the zone.
   ///   See [ContextScope] for detailed examples.
+  /// - [spanExceptionReporter]: Optional callback invoked when the body throws
+  ///   an exception. Use this to control how the error is recorded on the span
+  ///   (e.g., record a custom event, set custom attributes). When provided,
+  ///   the default `recordException` and `setStatus(error)` are skipped.
+  ///   **Important**: The exception is always rethrown regardless of this
+  ///   callback. Do NOT call `span.end()` inside this callback — the framework
+  ///   manages span lifecycle automatically.
   ///
   /// **Example - Basic usage:**
   /// ```dart
@@ -480,6 +487,22 @@ class Faro {
   /// rootSpan.end();
   /// ```
   ///
+  /// **Example - Custom exception reporting:**
+  /// ```dart
+  /// await Faro().startSpan(
+  ///   'payment',
+  ///   (span) async {
+  ///     span.setAttribute('payment.id', paymentId);
+  ///     await processPayment();
+  ///   },
+  ///   spanExceptionReporter: (span, error, stackTrace) {
+  ///     span.setStatus(SpanStatusCode.error, message: 'Payment failed');
+  ///     span.setAttribute('payment.error_code', getErrorCode(error));
+  ///     span.recordException(error, stackTrace: stackTrace);
+  ///   },
+  /// );
+  /// ```
+  ///
   /// See also:
   /// - [startSpanManual] for manual span lifecycle management
   /// - [ContextScope] for timer/stream context behavior
@@ -490,6 +513,7 @@ class Faro {
     Map<String, Object> attributes = const {},
     Span? parentSpan,
     ContextScope contextScope = ContextScope.callback,
+    SpanExceptionReporter? spanExceptionReporter,
   }) async {
     return _tracer.startSpan(
       name,
@@ -497,6 +521,7 @@ class Faro {
       attributes: attributes,
       parentSpan: parentSpan,
       contextScope: contextScope,
+      spanExceptionReporter: spanExceptionReporter,
     );
   }
 
--- a/lib/src/faro.dart
+++ b/lib/src/faro.dart
@@ -450,7 +450,9 @@ class Faro {
   ///   - [ContextScope.zone]: Stays active for timers/streams in the zone.
   ///   See [ContextScope] for detailed examples.
   /// - [exceptionOptions]: Controls how exceptions are recorded on the span.
-  ///   Overrides global [FaroConfig.spanExceptionOptions] for this span.
+  ///   Merged with global [FaroConfig.spanExceptionOptions]. Only explicitly
+  ///   set fields override the global values; omitted fields inherit from
+  ///   global config.
   ///   See [SpanExceptionOptions] for details.
   ///
   /// **Example - Basic usage:**
@@ -516,8 +518,9 @@ class Faro {
     ContextScope contextScope = ContextScope.callback,
     SpanExceptionOptions? exceptionOptions,
   }) async {
-    final effectiveOptions =
-        exceptionOptions ?? config?.spanExceptionOptions;
+    final effectiveOptions = (config?.spanExceptionOptions ??
+            SpanExceptionOptions.defaults)
+        .mergeWith(exceptionOptions);
     return _tracer.startSpan(
       name,
       body,
--- a/lib/src/faro.dart
+++ b/lib/src/faro.dart
@@ -21,6 +21,7 @@ import 'package:faro/src/session/session_id_provider.dart';
 import 'package:faro/src/session/session_sampling_provider.dart';
 import 'package:faro/src/tracing/faro_tracer.dart';
 import 'package:faro/src/tracing/span.dart';
+import 'package:faro/src/tracing/span_exception_options.dart';
 import 'package:faro/src/transport/batch_transport.dart';
 import 'package:faro/src/transport/faro_base_transport.dart';
 import 'package:faro/src/transport/faro_transport.dart';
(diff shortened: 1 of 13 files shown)


robert-northmind:
Hi @Rloaded , thanks for this contribution and the input 🤩 we really appreciate it ❤️

In general it looks great,and having this configuration possibility is very useful.

I just have some thoughts and feedback to how this could also be handled.

What about if we use an API that is a bit closer to common OTel SDK patterns.

Something like this:

```dart
FaroConfig(
  spanExceptionOptions: SpanExceptionOptions(
    recordException: true,
    setStatusOnException: true,
    exceptionSanitizer: mySanitizer,
  ),
);

await Faro().startSpan(
  'checkout',
  (span) async => processCheckout(),
  exceptionOptions: SpanExceptionOptions(
    recordException: true,
    setStatusOnException: true,
    exceptionSanitizer: (error, stackTrace) {
      return SanitizedSpanException(
        type: error.runtimeType.toString(),
        message: 'Checkout failed',
        stackTrace: null,
        statusDescription: 'Checkout failed',
      );
    },
  ),
);

```

I think something like this would give you all the flexibility you need.

You could configure it as global config, and per startSpan call.

recordException would control if the sdk should auto record exceptions on the span.

setStatusOnException would control if sdk should auto set the status to error.

exceptionSanitizer gives you a chance to customize how errors are recorded by the sdk.

So if you wanted to manage it all by yourself, you could set recordException=false and setStatusOnException=false. And then use the span in the startSpan scope to do what you want.

Would that work for your use case?
I think this fits a bit better to other existing Otel sdks for other languages.

lib/src/faro.dart
```diff
@@ -513,15 +514,17 @@ class Faro {
     Map<String, Object> attributes = const {},
     Span? parentSpan,
     ContextScope contextScope = ContextScope.callback,
-    SpanExceptionReporter? spanExceptionReporter,
+    SpanExceptionOptions? exceptionOptions,
   }) async {
+    final effectiveOptions =
+        exceptionOptions ?? config?.spanExceptionOptions;
```
robert-northmind: Nice, I like this approach. just some minor thoughts.
Here, we would override whatever global config we might have had. 
Like e.g. if we did:
```dart
FaroConfig(
  spanExceptionOptions: SpanExceptionOptions(
    exceptionSanitizer: sanitize,
  ),
);
```
then here in the startSpan we do
```dart
Faro().startSpan(
  'checkout',
  (_) async => checkout(),
  exceptionOptions: const SpanExceptionOptions(recordException: false),
);
```

we would loose that global `exceptionSanitizer`.
What do you think about doing some merging here instead.

One concept could be to have a `SpanExceptionOptions.defaults`.
so that the config.spanExceptionOptions always have a value.
and then here we could call like:

```dart
return _tracer.startSpan(
      name,
      body,
      attributes: attributes,
      parentSpan: parentSpan,
      contextScope: contextScope,
      exceptionOptions: effectiveOptions,
      exceptionOptions: config.spanExceptionOptions.mergeWith(exceptionOptions),
);
```


The SpanExceptionsOptions could then looks like this:
```dart
/// Controls how exceptions thrown from the body of `startSpan` are recorded.
///
/// Can be configured globally via [FaroConfig.spanExceptionOptions]
/// or per-span via the `exceptionOptions` parameter of `startSpan()`.
/// Per-span options are merged over global configuration.
class SpanExceptionOptions {
  /// Creates span exception options.
  ///
  /// Omitted parameters inherit from global configuration when these options
  /// are used as a per-span override. Without global configuration, omitted
  /// parameters default to the SDK's standard behavior:
  /// - [recordException]: `true` — auto-record exceptions on the span
  /// - [setStatusOnException]: `true` — auto-set span status to error
  /// - [exceptionSanitizer]: `null` — record the raw exception as-is
  const SpanExceptionOptions({
    bool? recordException,
    bool? setStatusOnException,
    this.exceptionSanitizer,
  }) : _recordException = recordException,
       _setStatusOnException = setStatusOnException;

  /// SDK default span exception behavior.
  static const defaults = SpanExceptionOptions(
    recordException: true,
    setStatusOnException: true,
  );

  final bool? _recordException;
  final bool? _setStatusOnException;

  /// Whether the SDK should automatically record the exception on the span.
  ///
  /// When `true` (default), the SDK calls `span.recordException()`.
  /// When `false`, the SDK skips automatic exception recording.
  bool get recordException => _recordException ?? true;

  /// Whether the SDK should automatically set the span status to error.
  ///
  /// When `true` (default), the SDK calls
  /// `span.setStatus(SpanStatusCode.error)`.
  /// When `false`, the SDK skips automatic status updates on exception.
  bool get setStatusOnException => _setStatusOnException ?? true;

  /// Optional callback to sanitize exception data before recording.
  ///
  /// When provided, the SDK uses the returned [SanitizedSpanException]
  /// to record the exception instead of the raw error object. This is
  /// useful for removing PII or sensitive data from error messages.
  ///
  /// The sanitizer is only invoked when [recordException] or
  /// [setStatusOnException] is `true`.
  final ExceptionSanitizer? exceptionSanitizer;

  /// Returns a new options object with [overrides] applied field-by-field.
  ///
  /// This allows a per-span override like
  /// `SpanExceptionOptions(recordException: false)` to keep a globally
  /// configured [exceptionSanitizer] while changing only one flag.
  SpanExceptionOptions mergeWith(SpanExceptionOptions? overrides) {
    if (overrides == null) {
      return this;
    }

    return SpanExceptionOptions(
      recordException: overrides._recordException ?? _recordException,
      setStatusOnException:
          overrides._setStatusOnException ?? _setStatusOnException,
      exceptionSanitizer: overrides.exceptionSanitizer ?? exceptionSanitizer,
    );
  }
}
```

lib/src/tracing/faro_zone_span_manager.dart
```diff
@@ -89,17 +90,43 @@ class FaroZoneSpanManager {
...
+                'exception.type': sanitized.type,
+                'exception.message': sanitized.message,
+                if (sanitized.stackTrace != null)
+                  'exception.stacktrace':
+                      sanitized.stackTrace.toString(),
+              });
+            }
           } catch (_) {
-            // Preserve original exception — don't let callback errors mask it
+            // Preserve original exception if sanitizer fails
```
robert-northmind: I think there is a risk here that the `final result = await body(span);` trows, and then also the `final sanitized = sanitizer(error, stackTrace);` could throw,
in that case, we end up in the `// Preserve original exception if sanitizer fails` and all is good. 

but then the span would not report the error. So looking at the span later, it would still be status=unset (which sort of implies: not failed).

Should we add this in here:
```dart
if (shouldSetStatus && !span.statusHasBeenSet) {
    span.setStatus(SpanStatusCode.error);
}
```

test/src/tracing/faro_zone_span_manager_test.dart
```diff
@@ -639,57 +690,111 @@ void main() {
...
+              SpanStatusCode.error,
+              message: 'Sanitized error',
+            ),
+          ).called(1);
+          verify(
+            () => mockSpan.addEvent(
+              'exception',
+              attributes: any(named: 'attributes'),
+            ),
+          ).called(1);
```
robert-northmind: Should we here also verify the data in the attributes. that they are correct. something like:

```dart
final attributes = verify(
  () => mockSpan.addEvent(
    'exception',
    attributes: captureAny(named: 'attributes'),
  ),
).captured.single as Map<String, Object>;
expect(attributes['exception.type'], 'Exception');
expect(attributes['exception.message'], 'Operation failed');
expect(attributes, isNot(contains('exception.stacktrace')));
expect(attributes.toString(), isNot(contains('user@email.com')));
```

robert-northmind (commented):
Hi @Rloaded  this looks great. 🚀
I just added some small comments for potential improvements. After this I think it is merge ready. What do you think?

lib/src/tracing/faro_zone_span_manager.dart
Rloaded: I agree, but I think a generic message, like "exception sanitizer failed", would give users a hint why the span is set as failed

lib/src/faro.dart
Rloaded: I really like the idea of merge with 👍🏽 
Let me update the PR

lib/src/tracing/faro_zone_span_manager.dart
robert-northmind: nice, I like it 👍  Let's go with that

robert-northmind (dismissed):
LGTM 🚀 
@Rloaded  just fix the failing CI step. I think it's the formatting which is failing.
Then this PR is good to be merged

robert-northmind (approved):

PR #258, opened 2026-05-28
chore(ci): attest SLSA provenance for pub.dev archive

domasx2:
## Description

Inline the `dart-lang/setup-dart` reusable publish workflow so we can attach SLSA build provenance to the `faro` archive on pub.dev.

`dart pub publish` builds the package tarball in memory with no stable local path, so there's nothing on disk to attest. After publishing, this workflow polls pub.dev's API for the new version, downloads the canonical archive pub.dev serves, cross-checks the sha256 against the API's reported `archive_sha256`, and attests those exact bytes via `actions/attest-build-provenance@v4.1.0`. Consumers fetching from pub.dev can verify with `gh attestation verify <tarball> --repo grafana/faro-flutter-sdk`.

## Type of Change

- [x] 🧹 Chore / Housekeeping

## Checklist

## Additional Notes

Tradeoffs:

- We now own the publish flow (~6 steps copied from upstream). Future `dart-lang/setup-dart` changes to the reusable workflow won't carry over automatically.
- If `dart pub publish` succeeds but the post-publish attestation block fails (pub.dev API hiccup, polling exhausted), the package is **published but unattested**. Recoverable by re-running the attestation steps manually.

_PR description authored by Claude on Domas's behalf._


.github/workflows/publish.yml
```diff
@@ -7,8 +7,94 @@ on:
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      # The dart-lang/setup-dart action provisions the OIDC token that
+      # `dart pub publish` uses for pub.dev's trusted-publisher flow.
+      - name: Setup Dart SDK
+        uses: dart-lang/setup-dart@e51d8e571e22473a2ddebf0ef8a2123f0ab2c02c # v1.7.2
```
robert-northmind: in the comment we say `v1.7.2`
but the hash actually refers to `v1.7.1`
the `v1.7.2` has this commit hash:
65eb853c7ba17dde3be364c3d2858773e7144260

https://github.com/dart-lang/setup-dart/tags

domasx2: Good catch — fixed in e694cf4. Bumped the pin to the actual `v1.7.2` commit (`65eb853c7ba17dde3be364c3d2858773e7144260`) so the SHA matches the comment.

_Reply posted by Claude on Domas's behalf._

.github/workflows/publish.yml
```diff
@@ -7,8 +7,103 @@ on:
...
+        env:
+          NAME: ${{ steps.pkg.outputs.name }}
+          VERSION: ${{ steps.pkg.outputs.version }}
+        run: |
+          # pub.dev is eventually consistent — poll until the new version is
+          # queryable. Typical propagation is well under a minute.
+          api_url="https://pub.dev/api/packages/${NAME}/versions/${VERSION}"
+          archive_url=""
+          expected_sha=""
+          for attempt in 1 2 3 4 5 6 7 8 9 10 11 12; do
```
aranhave: I have a small release-flow question. should we give pub.dev a bit longer than ~2 min here? Since publish is non-idempotent, if publish succeeds but this poll times out, rerunning the workflow would prob fail earlier at `dart pub publish`, leaving us with manual attestation only.

robert-northmind: Good point. I am giving it more time here. 10min

robert-northmind (dismissed):

.github/workflows/publish.yml
```diff
@@ -7,8 +7,104 @@ on:
...
+            archive_url=$(echo "$response" | jq -r '.archive_url // empty')
+            expected_sha=$(echo "$response" | jq -r '.archive_sha256 // empty')
+            if [ -n "$archive_url" ]; then
+              break
+            fi
+            echo "::notice::pub.dev has not surfaced ${NAME}@${VERSION} yet (attempt ${attempt}/60); sleeping 10s"
+            sleep 10
+          done
+          if [ -z "$archive_url" ]; then
+            echo "::error::pub.dev did not surface ${NAME}@${VERSION} within ~10 minutes; published but unattested"
```
aranhave, line 96: Non-blocking potential follow-up thought, if publish succeeds but a later step fails, re-running the workflow would die at `dart pub publish` since the version is already live, so recovery from "published but unattested" is manual-only today. Maybe worth a follow-up issue for a codified recovery path (e.g. an attest-only `workflow_dispatch`)? Not needed for this PR.

.github/workflows/publish.yml
```diff
@@ -7,8 +7,104 @@ on:
...
+            exit 1
+          fi
+          curl -fsSL -o package.tar.gz "$archive_url"
+          actual_sha=$(sha256sum package.tar.gz | awk '{print $1}')
+          if [ "$expected_sha" != "$actual_sha" ]; then
```
aranhave, line 105: nit: if pub.dev ever surfaces `archive_url` without `archive_sha256`, `expected_sha` is empty and this fails with a confusing `sha256 mismatch: pub.dev reports ,` message. Worth an explicit empty check with its own error? Still fails closed today, so just a diagnosability thing.

aranhave (commented):
Nice 🚀 Left two small non-blocking comments. one follow-up thought on recovery, one nit. Happy for this to land either way :)

.github/workflows/publish.yml
robert-northmind: Fixed in 971c9fc by adding an explicit empty `archive_sha256` check before downloading and comparing the archive.

aranhave (approved):
Looks good from my side. Thanks for addressing the follow-up. 👍

PR #260, opened 2026-05-29
chore(security): harden Flutter and Yarn installs

robert-northmind:
## Summary

Hardens CI/tooling installs for the Faro Flutter SDK:

- pins the Flutter SDK version used by CI via root .fvmrc and .tool-versions
- adds a local setup-flutter composite action, following the IRM mobile app pattern
- replaces floating channel: stable workflow setup with the pinned Flutter version
- enables immutable Yarn installs in the webview demo
- adds a 7-day npm minimal age gate for the webview demo

## Why

Floating Flutter stable makes PR/release checks less reproducible because CI can pick up a newly published Flutter SDK without a repo change. The pinned setup keeps normal CI deterministic while still allowing us to update Flutter intentionally through reviewed PRs.

The current pin is Flutter 3.44.0, which is the latest stable release from the official Flutter release feed and was published on 2026-05-18, so it is past the 7-day cooldown window.

## Third-party GitHub Actions review

I also reviewed the current workflows from a supply-chain/security angle:

- Workflows pass zizmor.
- Actions are pinned to full commit SHAs.
- Checkout uses persist-credentials: false.
- Workflow permissions are tightly scoped.
- dart-lang/setup-dart publish workflow uses OIDC for pub.dev publishing, which is preferable to static credentials.
- subosito/flutter-action remains the Flutter installer, but now it is called through a local wrapper with an exact Flutter version instead of floating on stable.

Possible follow-up: add weekly Flutter version-check automation, similar to IRM mobile, to open PRs for new stable Flutter releases after a 7-day cooldown.

## Testing

- corepack yarn install --immutable
- corepack yarn build
- zizmor .github/workflows
- git diff --check


yduartep (approved):
LGTM

PR #263, opened 2026-06-04
feat: populate structured mobile meta fields

### LINKED ISSUE #221: Populate new Faro mobile payload fields (Device/OS meta, app.installationId, exception.fatal)
## Context

The Faro payload spec ([grafana/faro](https://github.com/grafana/faro)) recently added four structured fields to support mobile use cases. The Flutter SDK already collects this data but emits it through flat `session.attributes` keys (or not at all). This issue tracks populating the new first-class fields so mobile data renders correctly across the FE O11y stack and aligns with OTel semantic conventions.

**Upstream spec PRs:**
- [grafana/faro#72 — Device & OS meta types](https://github.com/grafana/faro/pull/72) (`Meta.device`, `Meta.os`)
- [grafana/faro#73 — App.installationId](https://github.com/grafana/faro/pull/73)
- [grafana/faro#71 — Exception.fatal](https://github.com/grafana/faro/pull/71)

### 1. `Meta.device` — structured Device object
Today device info is collected in `lib/src/device_info/device_info_provider.dart` and flattened into `session.attributes` with a `device_` prefix via `lib/src/device_info/session_attributes_provider.dart`.

- [ ] Add a `Device` model aligned with the Faro spec (`manufacturer`, `model`, `modelId`, `brand`, `architecture`, `type`) and OTel `device.*` semconv.
- [ ] Wire it into `Meta.device` at init.
- [ ] Stop duplicating the same values into `session.attributes` once the collector accepts the structured field.

### 2. `Meta.os` — structured OS object
Same story — OS fields live flat in session attributes today.

- [ ] Add an `Os` model (`name`, `version`, `type`, `buildId`, `description`) aligned with OTel `os.*` semconv.
- [ ] Wire it into `Meta.os`.
- [ ] Remove the flat `device_os*` session attributes once parity is confirmed.

### 3. `App.installationId`
The SDK already generates a stable UUID per install in `lib/src/device_info/device_id_provider.dart`, currently exposed as `session.attributes.device_id`. The upstream spec PR notes that the Flutter SDK already populates this value — this issue is about surfacing it in the new structured field.

- [ ] Add `installationId` to the `App` model in `lib/src/models/app.dart`.
- [ ] Populate it from the existing persisted UUID.
- [ ] Keep `session.attributes.device_id` for one release for backward compatibility, then remove.

### 4. `Exception.fatal`
Today fatality is inferred from the `type` string (`crash` = native, `flutter_error` = Dart). ANR path also uses `flutter_error`, which is imprecise (`lib/src/integrations/native_integration.dart`).

- [ ] Add `fatal` bool to `FaroException` in `lib/src/models/exception.dart`.
- [ ] Set `fatal: true` for native crashes (PLCrashReporter path, `lib/src/faro.dart`) and ANRs.
- [ ] Set `fatal: false` for Flutter `FlutterError` and uncaught zone errors.
- [ ] Keep existing `type` strings for backward compatibility.

## Acceptance criteria

- [ ] All four structured fields are emitted in payloads from a sample app.
- [ ] Values match or exceed what's currently in `session.attributes`.
- [ ] Unit tests cover each new field.
- [ ] Migration plan documented for deprecating the duplicated flat keys.

## Related

- Parent epic: grafana/app-o11y-kwl#2844
- Sibling issue: faro-react-native-sdk equivalent (to be created)

aranhave:
## Description

Populates the new structured Faro mobile payload fields from the existing Flutter SDK device/app/error sources:

- `meta.device`
- `meta.os`
- `meta.app.installationId`
- `exception.fatal`

The old flat `session.attributes` values are kept for migration compatibility. Once collector/plugin parity is confirmed and downstream queries have moved to the structured fields, the duplicated flat mobile device/OS fields can be removed in a follow-up release.

## Related Issue(s)

Fixes #221  
Related to grafana/app-o11y-kwl#3504

## Type of Change

- [ ] 📝 Documentation
- [ ] 🏗️ Code refactoring
- [ ] 🧹 Chore / Housekeeping

## Checklist

## Screenshots / Payload Proof

Captured from a real example-app run against a local collector:

Faro Flutter payload proof

## Tests

Validated locally with:

- `dart format --set-exit-if-changed lib test`
- `git diff --check`
- `flutter analyze`
- `flutter test` (`622` tests passed)

Test coverage added/updated for:

- structured `meta.device` / `meta.os` serialization using Faro spec field names
- `meta.app.installationId` serialization
- payload serialization with both structured fields and legacy `session.attributes`
- Android/iOS device info mapping, including iPad/iPad Simulator tablet detection
- `exception.fatal` defaulting to `false`
- native crash / explicit fatal error path setting `fatal = true`
- `setAppMeta` preserving the generated installation id

Sample app verification:

- Ran the Flutter example app on Android emulator against a local test collector.
- Captured 28 real Faro POST payloads.
- Verified outbound payloads include `meta.app.installationId`, `meta.device`, `meta.os`, and legacy flat `session.attributes.device_*` fields.
- Triggered the example app diagnostics flow with “Throw Exception” and verified the emitted exception payload includes `fatal: false`.

## Additional Notes

A few implementation choices worth calling out for review:

- `app.installationId` reuses the SDK’s existing persisted install/device id.
- Legacy flat `session.attributes` are intentionally kept for migration compatibility.
- Android `device.type` is omitted instead of guessed because the current source does not reliably distinguish phones from tablets.
- iOS `os.build_id` is omitted because `device_info_plus` does not expose the real iOS build id; sending the kernel string would be misleading.
- Android native crashes, iOS PLCrashReporter crashes, and ANRs set `exception.fatal = true`; handled Flutter/Dart errors remain non-fatal.

### DIFF
--- a/lib/src/device_info/device_info_provider.dart
+++ b/lib/src/device_info/device_info_provider.dart
@@ -23,11 +23,13 @@ class DeviceInfoProvider {
     var deviceOs = _platformInfoProvider.operatingSystem;
     var deviceOsVersion = _platformInfoProvider.operatingSystemVersion;
     var deviceOsDetail = 'unknown';
+    String? deviceOsBuildId;
     var deviceManufacturer = 'unknown';
     var deviceModel = 'unknown';
     var deviceModelName = 'unknown';
     var deviceBrand = 'unknown';
     var deviceIsPhysical = true;
+    String? deviceType;
 
     if (_platformInfoProvider.isAndroid) {
       final androidInfo = await _deviceInfoPlugin.androidInfo;
@@ -36,6 +38,7 @@ class DeviceInfoProvider {
 
       deviceOs = 'Android';
       deviceOsVersion = release;
+      deviceOsBuildId = androidInfo.id;
       deviceOsDetail = 'Android $release (SDK $sdkInt)';
       deviceManufacturer = androidInfo.manufacturer;
       deviceModel = androidInfo.model;
@@ -58,6 +61,9 @@ class DeviceInfoProvider {
       deviceModelName = iosInfo.modelName;
       deviceBrand = iosInfo.model;
       deviceIsPhysical = iosInfo.isPhysicalDevice;
+      deviceType = iosInfo.model.toLowerCase().contains('ipad')
+          ? 'tablet'
+          : 'mobile';
     }
 
     final deviceInfo = DeviceInfo(
@@ -70,6 +76,8 @@ class DeviceInfoProvider {
       deviceModelName: deviceModelName,
       deviceBrand: deviceBrand,
       deviceIsPhysical: deviceIsPhysical,
+      deviceOsBuildId: deviceOsBuildId,
+      deviceType: deviceType,
     );
     _deviceInfo = deviceInfo;
     return deviceInfo;
--- a/lib/src/faro.dart
+++ b/lib/src/faro.dart
@@ -155,10 +155,16 @@ class Faro {
     final attributesProvider = await SessionAttributesProviderFactory()
         .create();
     final customAttributes = optionsConfiguration.sessionAttributes ?? {};
-    final defaultAttributes = await attributesProvider.getAttributes();
+    final deviceInfo = await attributesProvider.getDeviceInfo();
+    final deviceId = await attributesProvider.getDeviceId();
+    final defaultAttributes = await attributesProvider.getAttributes(
+      deviceId: deviceId,
+      deviceInfo: deviceInfo,
+    );
     // Merge custom attributes first, then default attributes
     // Default attributes take precedence if there are conflicts
     meta.session?.attributes = {...customAttributes, ...defaultAttributes};
+    _setDeviceAndOsMeta(deviceInfo);
 
     _nativeChannel ??= FaroNativeMethods();
     config = optionsConfiguration;
@@ -184,6 +190,7 @@ class Faro {
       appEnv: optionsConfiguration.appEnv,
       appVersion: appVersion,
       namespace: optionsConfiguration.namespace ?? '',
+      installationId: '$deviceId',
     );
 
     // Make sampling decision (once per session)
@@ -268,18 +275,40 @@ class Faro {
     required String appEnv,
     required String appVersion,
     required String? namespace,
+    String? installationId,
   }) {
     final appMeta = App(
       name: appName,
       environment: appEnv,
       version: appVersion,
       namespace: namespace,
+      installationId: installationId ?? _instance.meta.app?.installationId,
     );
     _instance.meta = Meta.fromJson({
       ..._instance.meta.toJson(),
       'app': appMeta.toJson(),
     });
     _instance._batchTransport?.updatePayloadMeta(_instance.meta);
+  }
+
+  void _setDeviceAndOsMeta(DeviceInfo deviceInfo) {
+    _instance.meta = Meta.fromJson({
+      ..._instance.meta.toJson(),
+      'device': Device(
+        manufacturer: deviceInfo.deviceManufacturer,
+        modelIdentifier: deviceInfo.deviceModel,
+        modelName: deviceInfo.deviceModelName,
+        brand: deviceInfo.deviceBrand,
+        isPhysical: deviceInfo.deviceIsPhysical,
+        type: deviceInfo.deviceType,
+      ).toJson(),
+      'os': Os(
+        name: deviceInfo.deviceOs,
+        version: deviceInfo.deviceOsVersion,
+        buildId: deviceInfo.deviceOsBuildId,
+        detail: deviceInfo.deviceOsDetail,
+      ).toJson(),
+    });
   }
 
   void _tearDownForReset() {
@@ -410,6 +439,7 @@ class Faro {
     required String value,
     StackTrace? stacktrace,
     Map<String, String>? context,
+    bool fatal = false,
   }) {
     var parsedStackTrace = <String, dynamic>{};
     if (stacktrace != null) {
@@ -421,6 +451,7 @@ class Faro {
       value,
       parsedStackTrace,
       context: context,
+      fatal: fatal,
     );
     _telemetryRouter.ingest(TelemetryItem.fromException(faroException));
   }
@@ -807,6 +838,7 @@ class Faro {
             _instance.pushError(
               type: 'crash',
               value: '$reason , status: $status',
+              fatal: true,
               context: {
                 'description': description,
                 'stacktrace': stacktrace,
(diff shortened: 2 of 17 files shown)


benjoseph-grafana (approved):
LGTM!

yduartep (approved):
Looks good! Remember to create a follow up issue for when the plugin queries are updated with the new meta device and os info

lib/src/device_info/device_info_provider.dart
```diff
@@ -36,6 +38,7 @@ class DeviceInfoProvider {
 
       deviceOs = 'Android';
       deviceOsVersion = release;
+      deviceOsBuildId = androidInfo.id;
```
robert-northmind, line 41: Question about the iOS version of the `deviceOsBuildId`. 
I saw in the PR description, that it is left out since it is not exposed for iOS. Should we leave maybe a dedicated part in code here in the iOs part that says something like:

```
if (iOS) {
   ...
   // iOS does not expose Os build number.. something something..
   deviceOsBuildId = null;
}
```

lib/src/device_info/device_info_provider.dart
```diff
@@ -58,6 +61,9 @@ class DeviceInfoProvider {
       deviceModelName = iosInfo.modelName;
       deviceBrand = iosInfo.model;
       deviceIsPhysical = iosInfo.isPhysicalDevice;
+      deviceType = iosInfo.model.toLowerCase().contains('ipad')
+          ? 'tablet'
+          : 'mobile';
```
robert-northmind, line 66: Does Android not have a way to get out deviceType? Looks like we only set it for iOS.

lib/src/models/device_info.dart
```diff
@@ -10,6 +10,8 @@ class DeviceInfo {
     required this.deviceModelName,
     required this.deviceBrand,
     required this.deviceIsPhysical,
+    this.deviceOsBuildId,
+    this.deviceType,
```
robert-northmind, line 14: Does `Device` and `Os` sort of replace the `DeviceInfo` maybe? Should we just have replaced it? and used the `Device` and `Os` to populate the legacy session-attributes?

lib/src/models/device_info.dart
```diff
@@ -62,4 +67,7 @@ class DeviceInfo {
 
   /// Whether the device is a physical device or an emulator/simulator.
   final bool deviceIsPhysical;
+
+  /// Device form factor, when known.
+  final String? deviceType;
```
robert-northmind, line 72: I guess we sort of know these different types.
could have been an enum, with a custom toString method?
And if not, then maybe we could have at least defined some example values here. like "tablet" / "phone"?

lib/src/models/os.dart
```diff
@@ -0,0 +1,32 @@
...
+  Os({this.name, this.version, this.buildId, this.detail});
+
+  Os.fromJson(dynamic json) {
+    name = json['name'];
+    version = json['version'];
+    buildId = json['build_id'];
+    detail = json['detail'];
+  }
+
+  String? name;
```
robert-northmind, line 11: Some docs here would maybe have been nice.. Like some examples of how these values might look like? Similar to what we had in DeviceInfo?

lib/src/faro.dart
```diff
@@ -155,10 +155,16 @@ class Faro {
     final attributesProvider = await SessionAttributesProviderFactory()
         .create();
     final customAttributes = optionsConfiguration.sessionAttributes ?? {};
-    final defaultAttributes = await attributesProvider.getAttributes();
+    final deviceInfo = await attributesProvider.getDeviceInfo();
+    final deviceId = await attributesProvider.getDeviceId();
+    final defaultAttributes = await attributesProvider.getAttributes(
+      deviceId: deviceId,
+      deviceInfo: deviceInfo,
+    );
```
robert-northmind, line 163: this looks a bit strange,
so we lift out the deviceId and deviceInfo from the `attributesProvider`, and then we pass in those values into the `attributesProvider.getAttributes` again. But seems like the attributesProvider would know how to get these?

lib/src/faro.dart
```diff
@@ -184,6 +190,7 @@ class Faro {
       appEnv: optionsConfiguration.appEnv,
       appVersion: appVersion,
       namespace: optionsConfiguration.namespace ?? '',
+      installationId: '$deviceId',
```
robert-northmind, line 193: I guess we are renaming the `deviceId` to instead be called `installationId`, which is actually what it actually is.
But should we also just rename the variable and provider, to be called like `installationIdProvider`?

robert-northmind (commented):
Just left some minor comments

lib/src/device_info/device_info_provider.dart
aranhave: Yeah, 💯 . I left it implicit because deviceOsBuildId defaults to null but an explicit comment in the iOS branch would make the intent clearer. I’ll add that in a follow-up.

lib/src/device_info/device_info_provider.dart
aranhave: Yeah, I checked this while working on it. `device_info_plus` doesn’t expose a reliable Android phone/tablet type directly, so I avoided guessing and left it unset for Android. We could add it later using screen-size/window-class style detection if we want that field populated.

lib/src/models/device_info.dart
aranhave: I thought about that too. I kept `DeviceInfo` as the internal collected shape for now because it feeds both the new structured Device/Os meta and the legacy session attributes during the migration. Once we remove the flat attrs, I agree we can probably simplify this and make Device/Os the main shape.

lib/src/models/device_info.dart
aranhave: Good point. Since the Faro payload field is ultimately a string, I kept this simple but adding documented expected values here would help. I’ll add examples like mobile / tablet in the follow-up :)

lib/src/models/os.dart
aranhave: 💯 Agreed, that would make this easier to read :) . I’ll add docs/examples for the `Os` fields in the follow-up too

lib/src/faro.dart
aranhave: I did that so the structured meta and legacy session attributes are built from the exact same deviceId / deviceInfo instance and we avoid fetching twice but I do agree the API reads a bit awkwardly lol I’ll see if I can make that cleaner in the follow-up :)

lib/src/faro.dart
aranhave: Yeah, I think that naming would be clearer now. i kept DeviceIdProvider to avoid mixing the schema change with a broader rename,but since this value is really the install id, I can include a small rename/cleanup in the follow-up if it stays contained.

Issue #32, opened 2025-02-11
Rename DeviceID to installationID?


robert-northmind:
In [this issue](https://github.com/grafana/faro-flutter-sdk/issues/18) we added support for something we called `DeviceID`
But it is not really a deviceID. Since if you reinstall the app in the same device, it will get a new ID.
Maybe it makes sense to rename it to installationId?

PR #266, opened 2026-06-08
chore: clarify mobile meta follow-up cleanup

aranhave:
## Description

Follow-up to [#263](https://github.com/grafana/faro-flutter-sdk/pull/263) based on review comments.

Cleaned up the mobile meta follow-up by moving the install identifier internals to `InstallationId` / `InstallationIdProvider`, while keeping migration compatibility:

- persisted storage key stays `device_id`
- legacy flat `session.attributes['device_id']` stays unchanged
- exported `DeviceId` remains as a deprecated alias for now

Also kept the mobile meta docs/comments around structured meta and legacy session attributes clearer.

## Related Issue(s)

Related to [#263](https://github.com/grafana/faro-flutter-sdk/pull/263/files)  
Related to [#221](https://github.com/grafana/faro-flutter-sdk/issues/221?reload=1?reload=1)
Related to: [#32](https://github.com/grafana/faro-flutter-sdk/issues/32)

## Type of Change

- [x] 📝 Documentation
- [x] 🏗️ Code refactoring
- [x] 🧹 Chore / Housekeeping

## Checklist

N/A

## Additional Notes

Validated with `flutter analyze` and `flutter test`.


lib/src/models/device_id.dart
```diff
@@ -1,3 +1,7 @@
+/// Stable SDK-generated identifier for this app installation.
+///
+/// This is still named DeviceId for legacy API compatibility, but it is emitted
+/// as app.installationId in Faro payloads.
 class DeviceId {
```
robert-northmind: This is just the name of the class. We could rename this without having to deal with any legacy API compatibility right?
So we could already name it `InstallationId` or maybe `DeviceInstallationId`?

lib/src/device_info/session_attributes_provider.dart
```diff
@@ -21,14 +21,25 @@ class SessionAttributesProvider {
...
-  Future<Map<String, Object>> getAttributes({
-    DeviceId? deviceId,
-    DeviceInfo? deviceInfo,
-  }) async {
-    deviceId ??= await getDeviceId();
-    deviceInfo ??= await getDeviceInfo();
-
-    final attributes = <String, Object>{
+  Future<CollectedSessionAttributes> collectAttributes() async {
+    final installationId = await getDeviceId();
```
robert-northmind:
```suggestion
    final installationId = await getInstallationId();
```

lib/src/device_info/session_attributes_provider.dart
robert-northmind: I guess we could rename this also to be called `InstallationIdProvider` or `DeviceInstallationIdProvider`

robert-northmind (commented):
Looks good 🤩 
I know we should support the legacy attributes for some time. so there will be some legacy code hanging around.
But if possible, it would be nice if we could keep the code as clean as possible, so like already moving away from the DeviceID naming pattern. Since moving forward we will call it InstallationId.

lib/src/device_info/device_id_provider.dart
```diff
@@ -9,6 +9,7 @@ class DeviceIdProvider {
   }) : _sharedPreferences = sharedPreferences,
        _uuidProvider = uuidProvider;
 
+  // Preserve the existing storage key so installation IDs survive SDK upgrades.
```
robert-northmind: Nice. Yeah I guess we are stuck with this ID here :)

lib/src/models/device_id.dart
aranhave: 👍 I went with `InstallationId` so it matches the Faro field name. I kept `DeviceId` only as a deprecated alias because it’s exported today.

lib/src/device_info/session_attributes_provider.dart
aranhave: 💯 renamed this to `InstallationIdProvider` and removed the old internal provider wrapper.

lib/src/device_info/device_id_provider.dart
aranhave: yup exactly 🙂 keeping the stored `device_id` key so existing installs keep the same stable ID after upgrade.

aranhave:
> I know we should support the legacy attributes for some time. so there will be some legacy code hanging around.
> But if possible, it would be nice if we could keep the code as clean as possible, so like already moving away from the DeviceID naming pattern. Since moving forward we will call it InstallationId.

Makes sense 👍 . I moved the internal path to InstallationId / InstallationIdProvider now and removed the old internal `DeviceIdProvider`. I kept `DeviceId` as a deprecated alias since it’s exported from the public barrel today but happy to drop it too if you’d prefer the clean break. Storage key and flat `device_id` attribute stay unchanged for migration.

lib/src/models/device_id.dart
```diff
@@ -1,10 +1,4 @@
...
 
-  final String _value;
-
-  @override
-  String toString() {
-    return _value;
-  }
-}
+@Deprecated('Use InstallationId instead.')
+typedef DeviceId = InstallationId;
```
robert-northmind, line 4: 👍

robert-northmind (approved):
LGTM 🚀
I think we can also link this Issue as being solved by this PR:
https://github.com/grafana/faro-flutter-sdk/issues/32

Issue #271, opened 2026-06-11
fix(errors): report Flutter framework errors that have no stack trace instead of dropping them


robert-northmind:
## Summary

Flutter framework errors without a stack trace are silently dropped instead of being reported.

## Details

`FlutterErrorIntegration` (`lib/src/integrations/flutter_error_integration.dart`) only pushes an error when `details.stack != null`. Framework errors that arrive without a stack trace are skipped entirely, so they never reach the `exceptions` array.

## Suggested action

Report the error without a `stacktrace` (the value/type are still useful) rather than dropping it. Minor, but it silently loses signal.

## Context

Found while documenting error/crash/ANR categorization (#26).

Issue #270, opened 2026-06-11
fix(anr): deduplicate ANRs reported by both the watchdog and ApplicationExitInfo


robert-northmind:
## Summary

A single ANR can be reported twice with no deduplication.

## Details

There are two independent ANR paths:

1. **Live watchdog** (`android/.../ANRTracker.java`, `lib/src/integrations/native_integration.dart`) — detects a blocked main thread while the app runs and reports it immediately (currently as `type: 'flutter_error'`, plus an `anr` measurement).
2. **`ApplicationExitInfo`** (`lib/src/faro.dart` `enableCrashReporter`) — on the next launch, an exit with `REASON_ANR` is reported (currently as `type: 'crash'`).

When the OS kills an app for an ANR that the watchdog already caught, the same ANR is reported once live and once on next launch, with no correlation or deduplication between the two.

## Suggested action

Decide on intended behavior and either deduplicate (e.g. correlate by timestamp/session) or clearly document that the two represent different observations (detected-while-alive vs killed-by-OS). Best handled together with the ANR type-alignment work in #184.

## Related

- #184 (align ANR/error types)
- #212 (structured ANR frames)

## Context

Found while documenting error/crash/ANR categorization (#26).

Issue #269, opened 2026-06-11
fix(ios): route native crash reports through the normal SDK transport


robert-northmind:
## Summary

iOS crash reports bypass the normal SDK transport, sending a raw HTTP POST with only partial metadata.

## Details

`CrashReportingIntegration.sendCrashReport` / `sendPostRequest` (`ios/faro/Sources/faro/CrashReportingIntegration.swift`) posts the crash directly to the collector with only `app` and `session` meta captured at init time. It does **not** go through `BatchTransport`, so it skips:

- batching and retry
- the data-collection policy and sampling
- current user meta and view meta
- any custom transports configured by the app

It also types the exception with the raw signal name (e.g. `SIGSEGV`) rather than `crash`.

Note: #220 addressed a separate iOS crash bug (stack frames were stripped to empty). This issue is the larger follow-up explicitly left out of #220's scope — routing iOS crashes through the same path as the rest of the SDK.

## Suggested action

Report iOS crashes on the next launch through the normal exception/transport path (as Android does via `ApplicationExitInfo`), with `type: 'crash'` and full current meta, so they benefit from batching, sampling, the data-collection policy, and custom transports.

## Related

- #220 (iOS crash stack frames — narrower fix)
- #151 (crash/ANR session attribution)

## Context

Found while documenting error/crash/ANR categorization (#26).

robert-northmind:
Confirmed on a real device that this is the blocker for end-to-end iOS crash delivery.

**Setup:** example app (release build, no debugger) on iPhone 15 Pro / iOS 26.5, native crash triggered via a force-unwrap of nil, app relaunched. Collector pointed at a local capture stub to observe exactly what the SDK emits.

**Result:** the SDK *does* capture and POST the crash. The captured request was:

- `POST /collect/<key>` with `Content-Type: application/json`
- header `x-api-token: <key>` (the normal transport uses `x-api-key`)
- body envelope `{ "exceptions": [ { type, value, stacktrace.frames... } ], "meta": { "session", "app" } }`
- the exception carried **18 stack frames** (after #220 fixes the empty-frames bug)

So capture works and the payload is well-formed against a permissive endpoint, but the **same crash never appears in the Grafana dev stack** — i.e. the real collector does not ingest this out-of-band POST. Prime suspects are the `x-api-token` header (vs `x-api-key`) and/or the minimal envelope/path, plus there is no batching/retry/offline so a single failed POST is lost silently.

[...]

Issue #268, opened 2026-06-11
chore(android): remove or complete dead JVM crash pipeline (ExceptionHandler/RumCache/lastCrashReport)


robert-northmind:
## Summary

The Android JVM uncaught-exception crash pipeline is dead code: it captures nothing and is never invoked.

## Details

The intended flow is `Thread.UncaughtExceptionHandler` (`android/.../ExceptionHandler.java`) → persist to `android/.../RumCache.java` → read back on next launch and surface via the Dart `lastCrashReport` method-channel handler (`lib/src/integrations/native_integration.dart`). In practice:

- `RumCache.setContext` is never called, so the cache has no context and writes fail.
- Nothing ever reads the cache back.
- No native code invokes the Dart `lastCrashReport` handler, so it is unreachable.
- Additionally, if it *were* invoked, `lastCrashReport` would push the crash as an error-level **log**, not a fatal exception — so it would also misclassify the signal.

Net effect: this path contributes nothing today; native crashes on Android are delivered (when they are) via the `ApplicationExitInfo` path in `enableCrashReporter`, not this one.

## Suggested action

Either remove `ExceptionHandler` / `RumCache` / the `lastCrashReport` handler entirely, or finish the pipeline so it emits a fatal **exception** (not a log). Removal is likely the cleaner choice given the `ApplicationExitInfo` path already exists.

## Context

Found while documenting error/crash/ANR categorization (#26).

PR #272, opened 2026-06-11
docs(errors): add error, crash, and ANR categorization reference

### LINKED ISSUE #26: Investigate what the differences are between errors and app-crashes
Which events leads to which error type?
Are we maybe wrongly reporting crashes which are actually just errors?
are log-errors showing up as errors? should they?

Image

robert-northmind:
## Description

Adds `doc/error-categorization.md`, a standalone reference describing how the
Faro Flutter SDK captures error-like signals and which category each one is
reported as.

The doc covers:

- The two signal kinds the SDK sends — **exceptions** (`pushError`) vs **logs**
  (`pushLog`) — and why an error-level log is not the same as an exception.
- The four categories: crash, ANR, non-fatal exception, and log.
- A table mapping each signal source (framework errors, async/zone errors,
  `ApplicationExitInfo`, the ANR watchdog, iOS PLCrashReporter, manual
  `pushError`/`pushLog`) to its category, `type`, and `fatal` flag.
- Guidance on choosing the right API.

It is written to stand on its own for SDK users — it does not reference
internal/private repos or the originating investigation issue.

While writing it, the investigation surfaced several pre-existing bugs/gaps
that did not have issues yet. These were filed separately rather than
documented here:

- #268 — dead Android JVM crash pipeline (`ExceptionHandler` / `RumCache` /
  `lastCrashReport` is never invoked)
- #269 — iOS crash reports bypass the normal SDK transport
- #270 — duplicate ANR reporting (live watchdog and `ApplicationExitInfo`)
- #271 — Flutter framework errors without a stack trace are silently dropped

## Related Issue(s)

Closes #26

## Type of Change

- [x] 📝 Documentation

## Checklist

- [ ] I have updated the CHANGELOG.md under the "Unreleased" section

(No tests or CHANGELOG entry: documentation-only change, no code or
user-facing behavior affected.)

## Additional Notes

The four follow-up issues above (#268, #269, #270, #271) were all opened from
this same investigation and can be triaged independently of this PR.

### DIFF
--- a/doc/error-categorization.md
+++ b/doc/error-categorization.md
@@ -1,0 +1,88 @@
+# Errors, Crashes, and ANRs
+
+This document explains how the Faro Flutter SDK captures error-like signals,
+which category each one is reported as, and how to choose the right API when
+you report something yourself.
+
+It is a reference for the categories the SDK produces today. File paths are
+relative to the repository root.
+
+---
+
+## Two signal kinds: exceptions and logs
+
+The SDK reports error-like information as one of two distinct signal kinds,
+carried in separate arrays of the Faro payload (`lib/src/models/payload.dart`):
+
+- **Exceptions** — `Faro().pushError(...)` creates a `FaroException`
+  (`lib/src/models/exception.dart`) with a `type`, `value`,
+  structured `stacktrace.frames`, an optional string `context` map, and a
+  `fatal` flag. It is serialized into the **`exceptions`** array.
+- **Logs** — `Faro().pushLog(..., level: LogLevel.error)` creates a `FaroLog`
+  (`lib/src/models/log.dart`) and is serialized into the **`logs`** array
+  with `level=error`.
+
+These stay separate end to end. A Faro backend materializes exceptions and
+logs as different record kinds (for example, `kind=exception` vs `kind=log`),
+and error/crash dashboards are typically built on the exception kind. So an
+**error-level log is not the same as an exception** and is generally not
+counted toward error or crash rates. If something represents a crash or an
+unhandled error, report it through `pushError` so it lands in the
+`exceptions` array; use `pushLog(level: LogLevel.error)` for diagnostic
+messages that should not inflate error counts.
+
+> The exact record kinds and field names you query depend on your Faro
+> backend (for example, Grafana Cloud Frontend Observability, or a
+> self-hosted collector). The two-array distinction above is part of the
+> Faro payload the SDK sends and is independent of where you send it.
+
+---
+
+## Categories
+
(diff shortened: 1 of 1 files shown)


doc/error-categorization.md
```diff
@@ -0,0 +1,88 @@
...
+4. **Log** — a message recorded via `pushLog`, including at `error` level.
+   Not a crash or an exception.
+
+---
+
+## How sources map to categories today
+
+| Signal source | Where | Reported as | `type` | `fatal` |
+|---------------|-------|-------------|--------|---------|
+| Flutter framework errors (`FlutterError.onError`) | `lib/src/integrations/flutter_error_integration.dart` | non-fatal exception | `flutter_error` | `false` |
```
aranhave, line 60: Should this row mention that `FlutterError.onError` only reports when `details.stack != null` today? Since #271 was filed for the no-stack case, this reads a bit like all framework errors are emitted.

doc/error-categorization.md
```diff
@@ -0,0 +1,88 @@
...
+| Manual `Faro().pushLog(..., level: LogLevel.error)` | `lib/src/faro.dart` | log | — | — |
+
+Notes:
+
+- Automatic Dart error capture (the first three rows) is controlled by
+  `enableErrorReporting`. Native crash and ANR capture is controlled by
+  `enableCrashReporting` / `anrTracking`.
+- Dart-level errors currently share the generic `type: 'flutter_error'`
+  rather than the underlying error class.
+- Native crash and ANR signals are detected from the **previous** run and
```
aranhave: is this too broad for ANRs? The runtime ANR watchdog is detected/polled while the app is running, while previous-run reporting applies to crashes and `ApplicationExitInfo` exits. Maybe worth narrowing this note so it does not imply every ANR is next-launch only?

aranhave (commented):
Nice doc, reads clearly and makes the current behavior much easier to follow :) . Left two tiny wording questions. 👍

doc/error-categorization.md
robert-northmind: Good catch, added a note that `FlutterError.onError` only reports when `details.stack != null` today, and pointed to #271 for the no-stack case. (b2fbe73)

doc/error-categorization.md
robert-northmind: Agreed, narrowed it: the next-launch note now applies to native crashes and `ApplicationExitInfo` exits (including ANR exits), and calls out that the runtime ANR watchdog detects in-session instead. (b2fbe73)

aranhave (approved):
⭐

Issue #265, opened 2026-06-08
[FEATURE] Support Flutter debug symbol upload for stack trace deobfuscation


nessux:
## Feature Description

Add support for uploading Dart debug symbol files (.symbols) - and optimally iOS .dSYM bundles and Android ProGuard/R8 mapping files - so that obfuscated Flutter stack traces captured by the Faro SDK can be symbolicated server-side in Grafana Cloud.

## Problem/Use Case

Flutter apps built for release with `--obfuscate --split-debug-info` produce unreadable stack traces. The Faro Flutter SDK captures and ships these obfuscated frames as-is. There is currently no way to deobfuscate them within the Grafana/Faro ecosystem (only option is manual deobfuscation via CLI). This makes error tracking in production Flutter apps significantly less useful compared to competing platforms.

## Proposed Solution

- A CLI tool or upload API endpoint that accepts Dart .symbols files
- Server-side symbolication of incoming Dart stack traces

## Alternative Solutions

- Manual upload on a cloud frontend as an interim solution

## Platform Considerations

- [ ] This feature applies to iOS
- [ ] This feature applies to Android
- [x] This feature is platform-agnostic

## Impact Assessment

- **Breaking change**: No
- **Performance impact**: Low (no runtime changes)
- **Maintenance complexity**: Medium (server-side symbol storage, symbolication)

## Additional Context

- [Flutter obfuscation docs](https://docs.flutter.dev/deployment/obfuscate)
- [Flutter deobfuscation in Firebase Crashlytics](https://firebase.google.com/docs/crashlytics/flutter/get-deobfuscated-reports)
- [Flutter deobfuscation in Sentry](https://docs.sentry.io/platforms/dart/guides/flutter/debug-symbols/)

## Checklist

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I have clearly described the problem and proposed solution
- [x] I have considered the impact on existing functionality

PR #273, opened 2026-06-11
fix(ios): preserve stack frames in native crash reports

### LINKED ISSUE #220: [BUG] iOS native crashes are never delivered — broken CrashReportingIntegration pipeline
## Summary

Native iOS crashes (Swift `fatalError`, force-unwrap of nil, etc.) are never
delivered as `type=crash` events when the app is relaunched. Observed in a
Flutter app using `faro: ^0.14.0` with `enableCrashReporting: true`. The
equivalent Android path (`ApplicationExitInfo` via `FaroPlugin.getCrashReport`)
does deliver crash events, so the problem is iOS-specific.

Independent of whether a crash is captured at all by PLCrashReporter, reading
`ios/faro/Sources/faro/CrashReportingIntegration.swift` surfaces several bugs
that would prevent a captured crash from being usable, and at least one that
would strip it of any value if it did arrive.

## Observed behavior

Tested on iOS Simulator (iPhone 17, iOS 26.4) with a Flutter app configured as:

```dart
faro.runApp(
  optionsConfiguration: FaroConfig(
    appName: 'quickpizza-flutter',
    enableCrashReporting: true,
    // ...
  ),
  appRunner: () { runApp(...); },
);
```

Repro:
1. Launch app, tap a button that triggers a native crash
   (`fatalError("...")` or force-unwrap of a nil `String?`).
2. Process dies as expected.
3. Relaunch the app.
4. Look for `{app_id=...} | logfmt | type="crash"` in Loki.

Result: **no `type=crash` event ever appears.** The post-crash session starts
normally and emits `session_start`, `view_changed`, etc., but nothing carrying
the crash data.

In the same sessions, `pushError` from Dart does work, so the transport itself
is healthy and the app key / collector URL are valid.

## Code-level findings (v0.14.0)

Reading `ios/faro/Sources/faro/CrashReportingIntegration.swift`:

### 1. `sanitized(stackFrames:)` always returns an empty array

```swift
private func sanitized(stackFrames: [StackFrame]) -> [StackFrame] {
    guard let _ = stackFrames.last else {
        return stackFrames
    }
    return []
    func asDictionary(){

    }
}
```

If the input is non-empty, this unconditionally returns `[]`. Called from
`CrashReportExporter.formattedStack` → every exported crash loses all stack
frames. Looks like an abandoned implementation: the inner `asDictionary()` is
unreachable and empty. This is very likely a committed WIP.

### 2. Crash payload bypasses the normal Faro transport

```swift
func sendCrashReport(crash: Dictionary<String,Any>, config: Dictionary<String, Any>) {
    let meta = [
        "app":     config["app"],
        "session": config["session"],
    ]
    var crashPayload: Dictionary<String,Any> = [:]
    crashPayload["exceptions"] = [crash]
    crashPayload["meta"] = meta
    sendPostRequest(
        collector: config["collectorUrl"] as! String,
        apiToken:  config["apiKey"] as! String,
        payload:   crashPayload
    ) { ... }
}
```

A raw POST with an ad-hoc envelope (`{exceptions, meta}`) is sent directly to
the collector URL, bypassing `FaroTransport` / the Dart-side batching. This
means:

- No buffering, no retry, no offline queuing (`OfflineTransport` won't see it).
- The envelope shape differs from what the Dart transport sends — I haven't
  confirmed whether the collector accepts both, but it's at least a candidate
  reason for silent drops.
- The HTTP header is `x-api-token`; `FaroTransport` uses `x-api-key`. Likely
  harmless for collectors that authenticate via the app key in the URL path,
  but inconsistent.
- The async `URLSession.dataTask` is fired, then immediately afterwards the
  caller runs `crashReporter.purgePendingCrashReport()`. If the POST fails or
  the app is killed again before it completes, the report is lost with no
  retry path.

### 3. Swallowing of load/parse errors

```swift
if crashReporter.hasPendingCrashReport() {
    do {
        let data = try crashReporter.loadPendingCrashReportDataAndReturnError()
        let report = try PLCrashReport(data: data)
        var crashReport = try CrashReport(from: report)
        // ... minify / export / send ...
    } catch let error {
        print("CrashReporter failed to load and parse with error: \(error)")
    }
}
crashReporter.purgePendingCrashReport()
```

Any thrown error is logged to stdout and then `purgePendingCrashReport()` is
called unconditionally. A broken payload or a parsing failure deletes the
evidence from disk without any telemetry signalling that it happened.

## What I did NOT verify (calling out uncertainty)

- Whether `PLCrashReporter` is capturing the crash at all on iOS 26.4
  Simulator. The SDK uses `.BSD` signal handler type; Swift `fatalError` and
  force-unwrap traps can bypass BSD handlers on simulator in some
  configurations. It's plausible (not proven) that on simulator the crash
  never reaches disk in the first place, which would make the bugs above
  irrelevant for this specific repro but still real bugs for on-device use.
- Whether the collector rejects/drops the ad-hoc envelope. I haven't tested
  ingestion of `{exceptions:[...], meta:{...}}` POSTs independently.

The most useful next verification would be: run on a real iOS device, trigger
the same crashes, and see if anything arrives. If it does, the bugs above
still apply (empty frames, no retry). If it doesn't, we also have a
capture-side simulator problem to split into its own issue.

## Suggested fix direction

Rather than patching the existing file piecemeal, I'd argue for a larger
rewrite along these lines:

1. Delete the ad-hoc `sendPostRequest` path. Route captured crashes through
   the same `FaroTransport` the rest of the SDK uses, so they benefit from
   batching, retries, and `OfflineTransport`.
2. Remove `sanitized(stackFrames:)` entirely — or make it a no-op — so exported
   stack frames are preserved.
3. Don't unconditionally purge on parse failure. Keep the data around for one
   retry, or at minimum emit a distinct telemetry event when dropping.
4. Add a small integration test that simulates a PLCrashReport-shaped dump on
   disk and asserts the SDK produces a payload with non-empty frames, correct
   envelope.

Note: session attribution (crash reported against the post-crash session
instead of the crashed one) is a known cross-platform issue tracked in #151
and intentionally left out of scope here.

## Environment

- `faro: 0.14.0`
- Flutter 3.35.x (Dart 3.11.0 stable)
- iOS Simulator: iPhone 17 / iOS 26.4 (arm64, Apple Silicon host)
- Collector: `https://faro-collector-dev-us-east-0.grafana-dev.net/collect/<appKey>`

## Related

- #78 — investigation of compatibility with multiple crash reporting SDKs.
- #149 — platform and runtime compatibility tracker.
- #151 — crash/ANR telemetry attributed to wrong session (out of scope here).
- #219 — Android native crash stacktraces not delivered (different surface,
  same overall "native crash pipelines are fragile" theme).

robert-northmind:
## Description

Fixes a bug where **every captured iOS native crash was exported with an empty
stack trace**.

`CrashReportExporter.sanitized(stackFrames:)` in `CrashReportingIntegration`
was a non-functional work-in-progress stub: for any non-empty input it
returned `[]` (and contained an unreachable, empty nested function). Since
`formattedStack(for:)` mapped `sanitized(stackFrames:)` to the exported
payload, every crash with frames lost all of them.

This PR removes the dead `sanitized` step (and its only call site), so frames
flow straight through to the exported crash payload. The catch block around
loading/parsing a pending crash report now logs a clearer message before the
report is purged.

It also adds a **"Native Crash" button** to the example app's App Diagnostics
page (force-unwrap of nil on iOS, null dereference on Android, via a method
channel) so the native crash pipeline can be exercised directly — this is how
the fix was validated (see below).

### Validation

Validated on a **real device** (iPhone 15 Pro, iOS 26.5, release build, no
debugger). Triggering the native crash and relaunching produced a crash
report carrying **18 stack frames** in `exceptions[0].stacktrace.frames`
(where the previous behavior produced 0). PLCrashReporter capture works, and
the frames are preserved by this fix.

End-to-end delivery to the collector is still blocked by a **separate**,
already-tracked issue (#269): iOS crashes are sent via an out-of-band POST
(`x-api-token` header + minimal envelope) that the collector does not ingest.
That is out of scope here; this PR makes the captured frames correct so that,
once #269 is addressed, crashes arrive with usable stack traces. Frames are
unsymbolicated raw addresses in release builds (dSYM upload is the separate
#265 work).

## Related Issue(s)

Closes #220
Related to #269, #265

## Type of Change

## Checklist

There is no Swift unit-test harness in the repo; the fix is verified by a
real-device crash capture (18 frames) and `flutter build ios` compilation.
The example-app crash trigger follows the existing (untested) App Diagnostics
demo pattern.

## Additional Notes

The example-app change is not part of the published `faro` package, so it is
intentionally not listed in the CHANGELOG.

## Screenshot of updated example app

IMG_9034

### DIFF
--- a/example/ios/Runner/AppDelegate.swift
+++ b/example/ios/Runner/AppDelegate.swift
@@ -3,6 +3,8 @@ import UIKit
 
 @main
 @objc class AppDelegate: FlutterAppDelegate, FlutterImplicitEngineDelegate {
+  private var crashChannel: FlutterMethodChannel?
+
   override func application(
     _ application: UIApplication,
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
@@ -12,5 +14,27 @@ import UIKit
 
   func didInitializeImplicitFlutterEngine(_ engineBridge: FlutterImplicitEngineBridge) {
     GeneratedPluginRegistrant.register(with: engineBridge.pluginRegistry)
+
+    if let registrar = engineBridge.pluginRegistry.registrar(forPlugin: "FaroExampleCrash") {
+      let channel = FlutterMethodChannel(
+        name: "faro_example/crash",
+        binaryMessenger: registrar.messenger()
+      )
+      channel.setMethodCallHandler { (call, _) in
+        if call.method == "crashNative" {
+          // Intentional native crash for validating crash reporting:
+          // force-unwrap a value that is nil at runtime.
+          let value = AppDelegate.alwaysNil()
+          print(value!)
+        }
+      }
+      self.crashChannel = channel
+    }
+  }
+
+  // Returns nil, but the compiler cannot prove it, so the force-unwrap at
+  // the call site traps at runtime instead of being a compile-time error.
+  private static func alwaysNil() -> String? {
+    return nil
   }
 }
--- a/example/ios/Runner/AppDelegate.swift
+++ b/example/ios/Runner/AppDelegate.swift
@@ -20,12 +20,14 @@ import UIKit
         name: "faro_example/crash",
         binaryMessenger: registrar.messenger()
       )
-      channel.setMethodCallHandler { (call, _) in
+      channel.setMethodCallHandler { (call, result) in
         if call.method == "crashNative" {
           // Intentional native crash for validating crash reporting:
           // force-unwrap a value that is nil at runtime.
           let value = AppDelegate.alwaysNil()
           print(value!)
+        } else {
+          result(FlutterMethodNotImplemented)
         }
       }
       self.crashChannel = channel
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   parsed into structured frames (e.g. sanitized, obfuscated, or free-form
   lines) are now kept as raw text in the frame's `function` field.
   ([#102](https://github.com/grafana/faro-flutter-sdk/issues/102))
+- iOS native crash reports no longer lose all stack frames: a broken
+  work-in-progress sanitization step in `CrashReportingIntegration`
+  unconditionally replaced non-empty stack traces with an empty array
+  before export. The dead sanitization has been removed so frames flow
+  through to the exported crash payload. Load/parse failures of pending
+  crash reports are now logged with a clearer message before the report
+  is purged. ([#220](https://github.com/grafana/faro-flutter-sdk/issues/220))
 
 ## [0.16.0] - 2026-05-11
 
--- a/example/android/app/src/main/java/com/example/rum_sdk_example/MainActivity.java
+++ b/example/android/app/src/main/java/com/example/rum_sdk_example/MainActivity.java
@@ -1,6 +1,27 @@
 package com.grafana.faro_example;
 
+import androidx.annotation.NonNull;
 import io.flutter.embedding.android.FlutterActivity;
+import io.flutter.embedding.engine.FlutterEngine;
+import io.flutter.plugin.common.MethodChannel;
 
 public class MainActivity extends FlutterActivity {
+    private static final String CHANNEL = "faro_example/crash";
+
+    @Override
+    public void configureFlutterEngine(@NonNull FlutterEngine flutterEngine) {
+        super.configureFlutterEngine(flutterEngine);
+        new MethodChannel(
+                flutterEngine.getDartExecutor().getBinaryMessenger(), CHANNEL)
+                .setMethodCallHandler((call, result) -> {
+                    if (call.method.equals("crashNative")) {
+                        // Intentional native crash for validating crash
+                        // reporting: dereference a null on the main thread.
+                        String value = null;
+                        value.length();
+                    } else {
+                        result.notImplemented();
+                    }
+                });
+    }
 }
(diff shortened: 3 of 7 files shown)


example/ios/Runner/AppDelegate.swift
```diff
@@ -12,5 +14,27 @@ import UIKit
 
   func didInitializeImplicitFlutterEngine(_ engineBridge: FlutterImplicitEngineBridge) {
     GeneratedPluginRegistrant.register(with: engineBridge.pluginRegistry)
+
+    if let registrar = engineBridge.pluginRegistry.registrar(forPlugin: "FaroExampleCrash") {
+      let channel = FlutterMethodChannel(
+        name: "faro_example/crash",
+        binaryMessenger: registrar.messenger()
+      )
+      channel.setMethodCallHandler { (call, _) in
```
aranhave: should the iOS handler call `result.notImplemented()` for non-`crashNative` methods, like Android does? Otherwise a typo/future method call could just wait forever.

aranhave (commented):
Looks great overall, nice fix and the real-device validation makes it easy to trust the stack-frame behavior now 🚀

example/ios/Runner/AppDelegate.swift
robert-northmind: Good catch — fixed in 51409a6. The iOS handler now calls `result(FlutterMethodNotImplemented)` for non-`crashNative` methods, matching Android, so an unknown method completes instead of hanging. Verified the example still compiles for iOS (`flutter build ios --no-codesign`).

aranhave (approved):

PR #274, opened 2026-06-11
fix(exceptions): preserve stack trace lines that do not match expected format

### LINKED ISSUE #102: [BUG] Stacktraces lose information when not in expected format
## Bug Description

if a stacktrace is not in the format that faro expects, the stacktraces lose all data

## Steps to Reproduce

1. Pass in a stack trace after sanitizing it with this function
```dart

  StackTrace sanitize(StackTrace? stackTrace) {
    final trace = stackTrace ?? StackTrace.current;

    final clean = trace.toString();

    const avoidLines = [
      'package:logger/',
      'package:flutter/',
      'package:bloc/',
      'package:couchsurfing_logger/src/logger_impl.dart',
      'package:provider/',
      'package:nested/',
      'package:stack_trace',
      'dart:async',
      '<asynchronous',
    ];

    Iterable<String> lines() sync* {
      for (final line in clean.split('\n')) {
        if (avoidLines.any(line.startsWith)) {
          continue;
        }

        final trimmed = line.trim();

        if (trimmed.isEmpty) continue;

        yield trimmed;
      }
    }

    return StackTrace.fromString(lines().join('\n'));
  }
```

## Expected Behavior

Whatever format I provide as a stacktrace should be accepted

## Actual Behavior

Stack traces lose all data

Image

## Environment

`flutter --version`
```bash
Flutter 3.35.7 • channel stable • https://github.com/flutter/flutter.git
Framework • revision adc9010625 (4 weeks ago) • 2025-10-21 14:16:03 -0400
Engine • hash 6b24e1b529bc46df7ff397667502719a2a8b6b72 (revision 035316565a) (30 days ago) • 2025-10-21 14:28:01.000Z
Tools • Dart 3.9.2 • DevTools 2.48.0
```

- **Flutter version**: [e.g. 3.16.0]
- **Dart version**: [e.g. 3.2.0]
- **Faro Flutter SDK version**: [e.g. 0.3.4]
- **Platform**: both

```lock
  faro:
    dependency: "direct main"
    description:
      path: "."
      ref: "v0.3.5"
      resolved-ref: ba3446f833a28610fac1a30b9dfd7e2be08d4778
      url: "https://github.com/grafana/faro-flutter-sdk.git"
    source: git
    version: "0.3.5"
```

## Checklist

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I have provided all the requested information above
- [ ] I have tested with the latest version of the SDK

(The version I am using is pinned, I can update to the latest if this has been fixed already)

robert-northmind:
## Description

Fixes a bug where passing `pushError` a stack trace whose lines don't match
the SDK's expected format caused **all** stack frame data to be silently lost.

`FaroException.stackTraceParse` split each line on spaces and indexed
`sf[sf.length - 2]`. A line with **no space** made that index `sf[-1]`, throwing
a `RangeError` that aborted the entire parse — and since `pushError` calls
`stackTraceParse` without a try/catch, the error propagated and the whole
exception was dropped. (`int.parse` on empty regex groups could also throw, and
non-matching lines became junk `null:null` frames.)

This came up for a user who sanitized their traces (filtering framework lines,
trimming) before reporting them — producing `stack_trace`-package-format lines
the parser didn't recognize.

### Fix

Parsing is now **per-line and total** via `_parseStackFrameLine`:

- Well-formed Dart frames produce **byte-for-byte identical output** to before
  (verified against the existing happy-path tests — no regression for normal
  traces).
- Any line that can't be parsed into structured fields is **preserved** (the
  raw line becomes the frame's function) instead of crashing the parse or being
  dropped. Blank lines are skipped.

So whatever format you provide, the information is kept rather than lost.

### Validation

- **Unit tests** (144 new lines): the reporter's exact sanitized-trace case
  plus obfuscated, async-marker, `stack_trace`-format, and no-space inputs.
- **End-to-end on the Android emulator + Grafana Cloud:** added a "Custom-Trace
  Error" button to the example app's App Diagnostics page that pushes an error
  with a non-standard trace (including the no-space line that previously
  destroyed the parse). The exception arrives in Grafana with **all four lines
  preserved as frames** — confirming the fix works through real ingestion, not
  just in unit tests.

## Related Issue(s)

Closes #102

## Type of Change

## Checklist

## Additional Notes

A pre-existing cosmetic quirk (the `package` module prefix renders as an empty
string before the filename's colon) is intentionally left untouched so the
happy-path output stays byte-for-byte identical; it can be a separate follow-up.

The example-app demo button is not part of the published `faro` package, so it
is intentionally not listed in the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   download the canonical archive pub.dev serves and attest those bytes via
   `actions/attest-build-provenance`. Consumers can verify with
   `gh attestation verify <tarball> --repo grafana/faro-flutter-sdk`.
+### Fixed
+
+- Preserve stack trace lines that do not match the expected Dart VM format
+  instead of silently dropping the whole stack trace. Lines that cannot be
+  parsed into structured frames (e.g. sanitized, obfuscated, or free-form
+  lines) are now kept as raw text in the frame's `function` field.
+  ([#102](https://github.com/grafana/faro-flutter-sdk/issues/102))
 
 ## [0.16.0] - 2026-05-11
 
--- a/example/lib/features/app_diagnostics/domain/app_diagnostics_demo_service.dart
+++ b/example/lib/features/app_diagnostics/domain/app_diagnostics_demo_service.dart
@@ -1,3 +1,4 @@
+import 'package:faro/faro.dart';
 import 'package:faro_example/shared/models/demo_log_entry.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -35,6 +35,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   download the canonical archive pub.dev serves and attest those bytes via
   `actions/attest-build-provenance`. Consumers can verify with
   `gh attestation verify <tarball> --repo grafana/faro-flutter-sdk`.
+### Fixed
```
aranhave, line 38: Tiny nit: should we add a blank line before this header, and maybe merge it with the other Unreleased sections?

aranhave (approved):
Looks good ✅ Just left a tiny optional nit

CHANGELOG.md
robert-northmind: thanks for fining this. I have a queue of like 4 PR coming through soon. And they will all touch the Changelog. So I can fix this in the next

PR #280, opened 2026-07-02
fix(offline): survive malformed cache data

### LINKED ISSUE #22: writeToFile in OfflineTransport is not handling invalid json
We are seeing a crash if the payload sent Into writeToFile is not valid jsonEncodable.
We should handle this and maybe at least try-catch it.

See here for more info: https://dash.embrace.io/app/5cCEk/v/2.2.16*2.2.17/session/last_48_hour/flutter_exception/84128A562BB9569699E31184EBDCC1A1

robert-northmind:
## Description

Hardens the offline transport cache against two failure modes:

- **Read path (primary):** A corrupt cached entry (unparseable payload or
  wrong-typed timestamp) is now skipped and purged instead of aborting the
  read. Previously a single bad line permanently poisoned the cache and
  blocked all remaining cached telemetry from being sent after reconnecting.
- **Write path (secondary):** Payloads whose user-supplied log `context` or
  event `attributes` cannot be JSON-encoded (a `DateTime`, custom object, or
  non-finite double) are dropped with a type-only diagnostic instead of
  failing the cache write. Diagnostics never include payload contents.

Note: `Measurement.values` and exception context are already sanitized
upstream, so only logs and events can reach the write guard. The
batch-level drop granularity (one non-encodable item drops its whole
batch) is being tracked as a separate follow-up for per-item sanitization
at ingest.

## Related Issue(s)

Fixes #22

## Type of Change

## Checklist

## Additional Notes

Hard to validate on-device, so verified via unit tests, `flutter analyze`,
and `dart format`. New tests in
`test/src/offline_transport/offline_transport_test.dart` cover both paths:
corrupt-line skipping, non-encodable payload drops (DateTime / NaN /
custom object), and retention of valid-but-unsent lines on send failure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Offline transport is now resilient to malformed cache data
+  ([#22](https://github.com/grafana/faro-flutter-sdk/issues/22)).
+  A corrupt cached entry (e.g. from a partial write or schema drift) is
+  now skipped and purged instead of aborting the read and permanently
+  blocking all remaining cached telemetry. Additionally, payloads whose
+  user-supplied log context or event attributes cannot be JSON-encoded
+  (a `DateTime`, custom object, or non-finite double) are dropped with a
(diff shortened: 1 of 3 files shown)


aranhave (approved):
LGTM!

PR #282, opened 2026-07-03
fix(offline): add timeout and robust error handling to connectivity probe

### LINKED ISSUE #11: Internet connectivity check in Offline_transport not reliable

robert-northmind:
## Description

The DNS probe used by `OfflineTransport` (via `InternetConnectivityService`) to
confirm real internet access had two reliability gaps:

- `InternetAddress.lookup` had **no timeout** — on some platforms it can hang for
  a long time, stalling online/offline decisions.
- Only `SocketException` was caught, letting other probe errors propagate.

This PR hardens the probe:

- Bounds the lookup with a 5s timeout (`defaultLookupTimeout`); a timed-out probe
  counts as a failed probe (⇒ offline).
- Catches `TimeoutException`, `SocketException`, and any other probe error.
- Keeps **probe-failure ⇒ offline** semantics deliberately: payloads cached while
  offline are flushed when connectivity returns, whereas a send attempted while
  actually offline is dropped by `FaroTransport` without retry. A false "offline"
  only costs disk usage; a false "online" risks permanent data loss.
- Makes the lookup an explicit **required** dependency wired via the factory
  (`InternetAddress.lookup`) and injectable for tests; adds optional
  `lookupTimeout`. Public API otherwise unchanged.

Validated on a real Android device (offline → cache → reconnect → flush, 0
payloads dropped) and cross-checked server-side in Loki. Unit tests cover the
timeout and error branches.

## Related Issue(s)

Related to #11

> Note: this **hardens** the existing probe; it does not implement the broader
> connectivity re-architecture #11 envisions. See the follow-up analysis I've
> added to #11 for the proposed "outcome-driven delivery" direction.

## Type of Change

## Checklist

- [x] I have made corresponding changes to the documentation (dartdoc added)
- [x] I have added tests that prove my fix is effective

## Additional Notes

This intentionally does **not** close #11. The full rework (outcome-driven
delivery, failure taxonomy, bounded cache) is captured as follow-up in the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   user-supplied log context or event attributes cannot be JSON-encoded
   (a `DateTime`, custom object, or non-finite double) are dropped with a
   type-only diagnostic rather than failing the cache write.
+- **Offline transport connectivity probe reliability**
+  ([#11](https://github.com/grafana/faro-flutter-sdk/issues/11)): The DNS
+  lookup used by `OfflineTransport` to confirm internet access now has a
+  5-second timeout (previously it could hang indefinitely on some
+  platforms, stalling online/offline decisions) and catches all probe
+  errors instead of only `SocketException`. Any probe failure is still
+  treated as offline: while offline, payloads are cached on disk and
+  flushed once connectivity returns, whereas a send attempted while
+  actually offline is dropped without retry — so a false "offline" only
+  costs disk usage, while a false "online" would risk permanent data
+  loss. The lookup function is also injectable for testing via the new
+  optional `addressLookup` and `lookupTimeout` parameters on
+  `InternetConnectivityService`.
 
 ## [0.17.0-beta.1] - 2026-07-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,8 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   actually offline is dropped without retry — so a false "offline" only
   costs disk usage, while a false "online" would risk permanent data
   loss. The lookup function is also injectable for testing via the new
-  optional `addressLookup` and `lookupTimeout` parameters on
-  `InternetConnectivityService`.
+  required `addressLookup` parameter and optional `lookupTimeout`
+  parameter on `InternetConnectivityService`.
 
 ## [0.17.0-beta.1] - 2026-07-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,9 +27,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   flushed once connectivity returns, whereas a send attempted while
   actually offline is dropped without retry — so a false "offline" only
   costs disk usage, while a false "online" would risk permanent data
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  lookup used by `OfflineTransport` to confirm internet access now has a
+  5-second timeout (previously it could hang indefinitely on some
+  platforms, stalling online/offline decisions) and catches all probe
+  errors instead of only `SocketException`. Any probe failure is still
+  treated as offline: while offline, payloads are cached on disk and
+  flushed once connectivity returns, whereas a send attempted while
+  actually offline is dropped without retry — so a false "offline" only
+  costs disk usage, while a false "online" would risk permanent data
+  loss. The lookup function is also injectable for testing via the new
+  optional `addressLookup` and `lookupTimeout` parameters on
```
aranhave: nit: `addressLookup` is required on the internal constructor, while only `lookupTimeout` is optional. Could we tweak this wording so it does not describe both parameters as optional?

aranhave (dismissed):
Looks good to me. 👍 Just left one non blocking comment

CHANGELOG.md
robert-northmind: nice, thanks. fixed it now

lib/src/offline_transport/internet_connectivity_service.dart
```diff
@@ -62,14 +92,30 @@ class InternetConnectivityService {
...
+  /// `OfflineTransport` and flushed when connectivity returns, whereas a
+  /// send attempted while actually offline is dropped by `FaroTransport`
+  /// without retry. A false "offline" therefore only costs disk usage,
+  /// while a false "online" risks permanent data loss.
   Future<bool> _isConnectedToInternet() async {
     try {
-      final result = await InternetAddress.lookup(
+      final result = await _addressLookup(
         _internetConnectionCheckerUrl,
-      );
```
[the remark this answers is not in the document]
robert-northmind: Good catch, thanks — fixed in a2c0547.

`_handleConnectivityResults` now registers a unique token per call (`_latestProbeToken`) before awaiting the DNS probe, and only applies its result if that token is still the latest when the probe returns; otherwise the stale result is discarded (last-started-wins). This prevents an older, slow/timed-out probe from overwriting the state a newer probe already resolved.

Added a regression test (`ignores a stale probe result that completes after a newer probe`) that drives the exact ordering: probe 1 hangs, probe 2 succeeds → online, then probe 1 completes as a failure — asserting `isOnline` stays `true`.

aranhave (approved):

Issue #284, opened 2026-07-07
[FEATURE] Re-evaluate session sampling on rotation (match Faro Web SDK)


robert-northmind:
## Feature Description

Re-evaluate the sampling decision each time the session rotates, so each
rotated session gets its own sampling roll — matching the Faro Web SDK.

## Problem/Use Case

Session rotation (#52) generates a new session id on inactivity/lifetime
timeout, but the sampling decision is made **once at init** and applied for the
whole app run. The Faro Web SDK instead re-samples per session, so a long-running
app produces independently-sampled sessions rather than one decision covering
every rotation.

## Proposed Solution

- On rotation, recompute the sampling decision (same sampler/rate used at init)
  for the new session.
- Apply the new decision to the transport pipeline so telemetry for the rotated
  session is included/excluded accordingly.

## Implementation Notes

- Transports are currently constructed once at init from the initial decision;
  runtime swapping of the active sampling decision needs supporting plumbing.
- The rotation hook already exists (`Faro._rotateSession`); the sampling
  re-evaluation would slot in there.

## Impact Assessment

- **Breaking change**: No
- **Performance impact**: Low
- **Maintenance complexity**: Medium (transport pipeline needs to support
  swapping the sampling decision at runtime)

## References

- Follow-up to #52 (session rotation)

## Checklist

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I have clearly described the problem and proposed solution
- [x] I have considered the impact on existing functionality

Issue #283, opened 2026-07-07
feat(session): persist minimal state and link cold-start sessions


robert-northmind:
## Summary

Align Faro Flutter session persistence with the shared Mobile O11y session contract.

Persist a minimal, versioned session record by default during Mobile O11y private preview. Every cold start creates a new session. When the previous session ID is known, the new session links to it instead of resuming the same live session across process starts.

Parent: https://github.com/grafana/frontend-o11y-knowledge-workbench/issues/57
Contract: https://github.com/grafana/frontend-o11y-knowledge-workbench/pull/97

## Required behavior

- Persist only the current session ID, previous session ID, session start time, last activity time, session sampling decision, and schema version.
- On a cold start without persisted state, create a new session normally.
- On a cold start with persisted state, create a new session and link the persisted session ID as the previous session.
- Do not continue the same live session across a process start.
- Enable this minimal persistence by default during Mobile O11y private preview.
- Do not persist user attributes or dynamic network, battery, or device state as part of the session record.

## Process and isolate ownership

A session may be shared only when the platform integration provides a multi-process-safe store and guarantees one writer at a time. SharedPreferences is not a multi-process-safe session store.

Without safe coordination:

- each process or background runtime keeps a separate session chain;
- each process links only to its own previous session when known;
- runtimes do not concurrently update one persisted session record; and
- telemetry includes process identity so independent chains can be correlated when needed.

## Edge cases

- corrupt or unsupported persisted records
- SDK storage migration
- app data being cleared
- overlapping foreground and background runtimes
- long gaps between launches

## Out of scope

These are tracked separately under the parent epic:

- the exact meaningful-activity allowlist
- the public reset API shape
- delayed-crash accounting before mobile billing
- per-session sampling behavior beyond persisting the current decision

## Acceptance criteria

- [ ] Persistence is enabled by default for the private-preview configuration.
- [ ] The persisted record is minimal and versioned.
- [ ] Every cold start creates a new session ID.
- [ ] A known previous session ID is linked on cold start.
- [ ] Persisted state never resumes the same live session across process starts.
- [ ] Unsupported or corrupt persisted state fails safely by starting a new session.
- [ ] Separate runtimes do not share unsafe concurrent writes.
- [ ] Process identity is included where separate session chains require correlation.
- [ ] Tests cover cold starts with and without state, migration/corruption, and supported concurrent-runtime behavior.

Issue #144, opened 2026-02-10
Session and Identity Lifecycle


robert-northmind:
## Summary

Align session and installation identity behavior with Faro expectations so
telemetry is correctly grouped over time and across app runtime contexts.

## Problem

Current session and identity handling appears to have gaps around:
- session inactivity timeout and max lifetime rotation
- consistency across normal app runtime vs background/runtime edge cases
- naming and semantics of installation identity
- lifecycle tracking around install/update behavior

These gaps can reduce data quality, correlation accuracy, and debuggability.

## Goals

- Define and implement clear session lifecycle rules.
- Ensure session rollover behavior is deterministic and observable.
- Clarify identity semantics (device vs installation identity).
- Improve lifecycle telemetry for install/update flows.

## Out of Scope

- General HTTP/network telemetry quality (tracked elsewhere).
- Offline transport reliability internals (tracked elsewhere).

## Acceptance Criteria

- [ ] Session expiration/rotation rules are documented and implemented.
- [ ] Session behavior is validated for foreground/background runtime paths.
- [ ] Identity field naming/semantics are finalized and documented.
- [ ] Install/update lifecycle behavior is defined and tracked.
- [ ] Sub-issues are linked and kept as execution units for this epic.

## Notes

This issue is intentionally public-safe and avoids internal implementation
details or private infrastructure context.

robert-northmind:
### Idle apps churn empty sessions because passive vitals force rotation without extending the window

Surfaced while validating session rotation (#52) on-device.

**Current behavior (correct and intended for #52):**
- Session *activity* = user/app-originated telemetry (events, logs, exceptions, app-developer measurements, user interactions, view changes, app lifecycle). Passive vitals (CPU, memory, refresh rate, frame stats, ANR count, app start) do **not** extend the inactivity window.
- However, expiry is checked on *every* ingested telemetry item, including vitals, so the triggering item always lands on a live session (an already-expired session would be dropped server-side).

**Observed consequence:**
- A foregrounded-but-idle app rotates its session roughly every 15 min purely from vitals: the first vitals push past the inactivity threshold rotates the session (`session_extend`, new session id, `previousSession` link), then repeats every threshold period. On Android the Dart isolate keeps running while the app is backgrounded, so this also happens continuously for an app sitting in the background → a long chain of short, activity-free sessions.

**Two options considered, and why neither is a clean fix on its own:**
[...]

PR #285, opened 2026-07-08
feat(session): rotate session on inactivity and max lifetime (#52)

### LINKED ISSUE #52: Handle Session inactivity and max lifetime
We currently create one sessionID for each app lifecycle.
But this is not correct. According to the Faro session rules, a session should get invalidated and refreshed if:

- there has been no activity for 15 min
- the max time of 4 hours has passed

We should add some logic into the Session handling to account for this.
And then, if we reach inactivity or max lifetime, we should make sure that we generate a new sessionID. 
**And probably also**, we should send a new event at this point with "session_started".

And maybe, I am not sure about this, but I think it makes sense, we should send the previous sessionID in as a Session-attribute as some sort of `parent-session-id`. Since it probably makes sense to link them together.

robert-northmind:
## Description

Implements automatic session rotation for the Faro Flutter SDK.

A session rotates when either window elapses:
- **15 minutes of inactivity**, or
- **4 hours of total lifetime**.

These windows are **fixed, not configurable** — the Grafana Cloud Faro
receiver enforces the same windows server-side and drops telemetry from
sessions that exceed them, so a longer client-side value would only cause
silent data loss.

**Session semantics:**
- The initial session emits `session_start`.
- A rotation emits `session_extend` and records the prior session id in the
  `previousSession` attribute, so consumers can stitch consecutive sessions
  together.
- A new session id is generated via the shared `SessionIdProvider`, so spans
  and other telemetry created after a rotation automatically carry the new id.

**What counts as "activity":**
- Telemetry from app or user behaviour extends the inactivity window: events,
  logs, exceptions, app-developer measurements, user interactions, view
  changes, app lifecycle events, and **spans** (each exported span emits a
  Faro event through the same path, so tracked HTTP requests and custom spans
  keep the session alive).
- Automatic vitals measurements pushed by the SDK itself (CPU, memory, refresh
  rate, frame stats, ANR count, app start) extend the window **only while the
  app is in the foreground**. This keeps a foregrounded-but-idle app (e.g. a
  user reading a screen) in a single session, while a backgrounded app's
  vitals cannot keep its session alive — the Dart isolate keeps running when
  backgrounded on Android, so counting background vitals would prevent expiry
  entirely.
- Expiry is always checked when any telemetry arrives, so the triggering item
  lands on a fresh session rather than an expired one.

Sampling is intentionally **not** re-evaluated on rotation in this PR (see
follow-ups below).

**Design / refactor:**
- `SessionManager` now owns session identity and lifecycle. It is inert until
  `start()` and notifies a `SessionChangedListener` on the initial session and
  each rotation; `Faro` reacts to those notifications to update payload meta
  and emit the lifecycle events. This follows the ownership model discussed in
  the OTel session-management proposal
  (open-telemetry/opentelemetry-specification#4198).
- `AppLifecycleService` tracks foreground/background state; `SessionActivityPolicy`
  is the single home for the "what counts as activity" rule (including the
  foreground gate).
- `SessionManager`, `AppLifecycleService`, `NativeIntegration`, the telemetry
  router and the tracer are now wired through dartyPod providers/scopes,
  replacing the previous static singletons and making them straightforward to
  override in tests.

## Related Issue(s)

Fixes #52
Related to #144 (Session and Identity Lifecycle epic)

## Type of Change

- [x] 📝 Documentation
- [x] 🏗️ Code refactoring
- [ ] 🧹 Chore / Housekeeping

## Checklist

## Additional Notes

**Reference alignment.** Session event naming (`session_start` / `session_extend`)
and the `previousSession` attribute follow the Grafana Cloud Faro session
definition; the manager's ownership/observer design is inspired by the OTel
session-management proposal (#4198).

**On-device validation.** Verified end-to-end on an Android emulator against a
dev collector, including an A/B comparison against `main`:
- Initial launch emitted `session_start` with no `previousSession`.
- After 15 minutes backgrounded, the session rotated automatically ~6s past
  the threshold, emitting `session_extend` with a new session id and
  `previousSession` pointing at the prior session. Preserved session
  attributes carried across.
- Post-rotation telemetry carried the new session id.
- Foreground gate confirmed: backgrounded vitals did **not** extend the
  session (so it could expire) but still served as the check that triggered
  rotation.
- Telemetry shapes (event names, log levels, measurement types, spans) matched
  `main`; the only additive signals on this branch were `session_extend` /
  `previousSession`.

**Follow-ups (tracked):**
- #283 — cross-restart / multi-process session persistence.
- #284 — re-sample per rotated session (match the Faro sampling model).
- #144 — pause vitals collection while backgrounded, so a backgrounded app
  stops emitting vitals entirely (foreground gating already prevents them from
  extending the session).

### DIFF
--- a/lib/src/faro.dart
+++ b/lib/src/faro.dart
@@ -364,7 +364,8 @@ class Faro {
       _didAttachUiActivityMonitor = false;
     }
     // Evict the per-init singletons (session manager, app lifecycle
-    // service) so the next init resolves fresh instances.
+    // service) so the next init resolves fresh instances. Disposable
+    // instances (e.g. NativeIntegration's vitals timer) are cleaned up here.
     pod.clearScope(faroInitScope);
     _isInitialized = false;
   }
--- a/lib/src/faro.dart
+++ b/lib/src/faro.dart
@@ -17,7 +17,10 @@ import 'package:faro/src/integrations/native_integration.dart';
 import 'package:faro/src/integrations/on_error_integration.dart';
 import 'package:faro/src/models/models.dart';
 import 'package:faro/src/native_platform_interaction/faro_native_methods.dart';
+import 'package:faro/src/session/app_lifecycle_service.dart';
+import 'package:faro/src/session/session_activity_kind.dart';
 import 'package:faro/src/session/session_id_provider.dart';
+import 'package:faro/src/session/session_manager.dart';
 import 'package:faro/src/session/session_sampling_provider.dart';
 import 'package:faro/src/tracing/faro_otel_bootstrap.dart';
 import 'package:faro/src/tracing/faro_tracer.dart';
@@ -92,10 +95,7 @@ class Faro {
   bool get isSampled => _isSampled;
 
   Meta meta = Meta(
-    session: Session(
-      SessionIdProviderFactory().create().sessionId,
-      attributes: {},
-    ),
+    session: Session(_sessionIdProvider.sessionId, attributes: {}),
     sdk: Sdk(FaroConstants.sdkName, FaroConstants.sdkVersion),
     app: App(name: '', environment: '', version: ''),
     view: ViewMeta('default'),
@@ -108,7 +108,13 @@ class Faro {
 
   FaroNativeMethods? get nativeChannel => _nativeChannel;
 
+  static SessionIdProvider get _sessionIdProvider =>
+      pod.resolve(sessionIdProviderProvider);
+
   TelemetryRouter get _telemetryRouter => pod.resolve(telemetryRouterProvider);
+
+  NativeIntegration get _nativeIntegration =>
+      pod.resolve(nativeIntegrationProvider);
 
   UserActionsService get _userActionsService =>
       pod.resolve(userActionsServiceProvider);
@@ -116,7 +122,7 @@ class Faro {
   UserActionUiActivityMonitor get _userActionUiActivityMonitor =>
       pod.resolve(userActionUiActivityMonitorProvider);
 
-  FaroTracer get _tracer => FaroTracerFactory().create();
+  FaroTracer get _tracer => pod.resolve(faroTracerProvider);
 
   @visibleForTesting
   set nativeChannel(FaroNativeMethods? nativeChannel) {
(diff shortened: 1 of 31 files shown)


lib/src/faro.dart
```diff
@@ -309,6 +324,34 @@ class Faro {
...
+    final attributes = <String, dynamic>{...?meta.session?.attributes};
+    if (previousId != null) {
+      attributes['previousSession'] = previousId;
+    }
+
+    meta = Meta.fromJson({
+      ...meta.toJson(),
+      'session': Session(currentId, attributes: attributes).toJson(),
+    });
+    _batchTransport?.updatePayloadMeta(meta);
```
aranhave, line 341: Could we also update the transport session header on rotation? The payload meta changes here, but the default `FaroTransport` still has the session id captured at init, so post-rotation requests can send a different `x-faro-session-id` than the body session id.

lib/src/integrations/native_integration.dart
```diff
@@ -200,3 +219,14 @@ class NativeIntegration {
...
 }
+
+/// Provides the [NativeIntegration].
+///
+/// Lives in [faroInitScope] so each `Faro.init` gets a fresh instance
+/// wired to the current telemetry router, and it is evicted by
+/// `Faro.resetForTesting`.
+final nativeIntegrationProvider = Provider<NativeIntegration>(
+  (pod) =>
+      NativeIntegration(telemetryRouter: pod.resolve(telemetryRouterProvider)),
```
aranhave, line 251: Since this is now scoped per init, should `NativeIntegration` keep and cancel the vitals timer on teardown? Otherwise the old timer can keep the old router/session manager alive after `resetForTesting` or re-init.

aranhave (commented):
Nice work overall 💯 . I left two questions around session rotation cleanup/header consistency.

lib/src/faro.dart
robert-northmind: good point :) I am looking into how to best handle this

lib/src/integrations/native_integration.dart
robert-northmind: good point here too. Looking into it.

robert-northmind: nice, great catch about this. I updated it now with a fix. and added a unit test to verify this also.

lib/src/faro.dart
robert-northmind: nice, fixed this now.

aranhave (approved):

PR #296, opened 2026-07-16
chore: prepare v0.17.0-beta.2

robert-northmind:
## Description

- Prepare the changelog and package metadata for `v0.17.0-beta.2`.
- Support explicit prerelease versions in `version_bump.dart`.
- Validate SemVer ordering and add regression tests.
- Document the prerelease release process.

## Related Issue(s)

N/A

## Type of Change

- [x] 📝 Documentation
- [ ] 🏗️ Code refactoring
- [x] 🧹 Chore / Housekeeping

## Checklist

- [x] I have added tests that prove my fix is effective
- [x] I have updated the CHANGELOG.md

N/A

## Additional Notes

Validation completed:

- `flutter test`
- `flutter analyze`
- `dart tool/pre_release_check.dart --post`

### DIFF
--- a/tool/version_bump.dart
+++ b/tool/version_bump.dart
@@ -1,28 +1,71 @@
 import 'dart:io';
 
+const _semVerPattern =
+    r'\d+\.\d+\.\d+(?:-[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?'
+    r'(?:\+[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?';
+
 /// Version bump types
 enum BumpType { patch, minor, major }
 
 /// Represents a semantic version
-class Version {
-  Version(this.major, this.minor, this.patch);
+class Version implements Comparable<Version> {
+  Version(
+    this.major,
+    this.minor,
+    this.patch, {
+    this.prerelease,
+    this.buildMetadata,
+  });
 
   /// Parse version string into Version object
   factory Version.parse(String version) {
-    final parts = version.split('.');
-    if (parts.length != 3) {
+    final match = RegExp('^($_semVerPattern)\$').firstMatch(version);
+    if (match == null) {
       throw FormatException('Invalid version format: $version');
     }
+
+    final coreAndPrerelease = version.split('+').first;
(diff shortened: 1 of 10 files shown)


tool/version_bump.dart
```diff
@@ -1,53 +1,39 @@
...
+const _semVerPattern =
+    r'\d+\.\d+\.\d+(?:-[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?'
+    r'(?:\+[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?';
+
+/// Resolve a bump type or explicit SemVer target into the next version.
+Version resolveNextVersion(String currentVersion, String target) {
+  final current = Version.parse(currentVersion);
+  final normalizedTarget = target.toLowerCase();
+  final next = switch (normalizedTarget) {
+    'patch' => Version(current.major, current.minor, current.patch + 1),
```
aranhave: Could we guard the bump keywords when the current version is a prerelease? `patch` from `0.17.0-beta.1` becomes `0.17.1`, skipping stable `0.17.0`, even though the release guide says prereleases should use an explicit target. Should this either reject bump keywords for prereleases or use the prerelease-aware `pub_semver` getters?

robert-northmind: hm.. good question. I guess we should still be allowed to bump even if last was a pre-release or? 
imagine we add some new logic which leads to a minor update. But good question. I will investigate it. thanks

robert-northmind: okay. I read up a bit on it and poked around a bit. And I think you are right. Best is to guard when the current is prerelease. Since it is not obvious what a bump would mean. I am updating that now. thanks :)

robert-northmind: Addressed in 6aa9d37. Bump keywords are now rejected when the current version is a prerelease, requiring an explicit target instead. Explicit targets remain validated with pub_semver and must be newer than the current version. Tests cover continuing the beta, promoting to stable, moving to the next minor line, rejecting all three shortcuts, and preventing equal or older targets.

domasx2 (approved):

aranhave (approved):

PR #299, opened 2026-07-24
feat(tracing): auto-correlate signals with active span context

robert-northmind:
## Description

Logs, events, exceptions, and measurements pushed via `pushLog`,
`pushEvent`, `pushError`, and `pushMeasurement` while a span is active are
now automatically stamped with the active span's `trace_id`/`span_id`,
matching the Faro Web SDK and the OpenTelemetry logs data model. This lets
you pivot from a log line straight to its trace in Grafana.

- New `FaroSpanContext(traceId:, spanId:)` value type and a
  `Span.spanContext` getter.
- All four push methods accept an optional `spanContext` override that
  takes precedence over the active span.
- Trace context is captured synchronously at push time, so buffered/offline
  signals keep the span that was active when they were recorded.
- Trace/span IDs come from the same source as OTLP span export, so log-side
  and Tempo-side IDs are produced identically and always match.

## Type of Change

- [x] New feature
- [x] Breaking change

## Breaking Changes & Migration

1. **Push method params (`trace` -> `spanContext`).**
   - `pushLog` / `pushEvent`: the existing `trace: Map<String, String>?`
     parameter is **replaced** by `spanContext: FaroSpanContext?`. This is
     the breaking part.
   - `pushError` / `pushMeasurement`: **newly gain** an optional
     `spanContext: FaroSpanContext?` (they had no trace parameter before, so
     this is purely additive).

   ```dart
   // Before
   Faro().pushLog('msg', level: LogLevel.info,
       trace: {'trace_id': t, 'span_id': s});
   // After
   Faro().pushLog('msg', level: LogLevel.info,
       spanContext: FaroSpanContext(traceId: t, spanId: s));
   ```

2. **HTTP integration exports.** `package:faro/faro.dart` now exports only
   `FaroHttpOverrides` from the HTTP integration. The implementation-detail
   wrappers `FaroHttpTrackingClient`, `FaroTrackingHttpClientRequest`, and
   `FaroTrackingHttpResponse` are no longer public.

3. **`FaroException.trace` type.** Changed from `String` to
   `Map<String, dynamic>?` (and now actually serializes). Code reading this
   field as a `String` must update.

Next release should be a minor bump (`0.18.0`) per pre-1.0 semver.

## Notes

- The HTTP network-error fallback log now carries its request span's context
  (previously none).
- Auto-captured Flutter errors (`flutter_error_integration` /
  `on_error_integration` / `run_zoned_integration`) flow through `pushError`,
  so they inherit whatever span is active in the zone at catch time. This is
  consistent with OTel semantics; usually desirable, but an unhandled error
  bubbling through an unrelated active span may be correlated to it.

## Test plan

- [x] `flutter analyze` clean
- [x] Full suite green (734 tests), incl. new integration test, model
  trace-serialization tests, and HTTP error-path test
- [x] Independent live end-to-end verification on Android emulator against
  Grafana Cloud dev: log/event/exception/measurement all correlated to the
  active span; Loki->Tempo click-through resolves; signals outside a span
  correctly carry no trace context


lib/faro.dart
```diff
@@ -7,10 +7,11 @@ export './src/faro.dart';
 export './src/faro_asset_tracking.dart';
 export './src/faro_navigation_observer.dart';
 export './src/faro_user_interaction_widget.dart';
-export './src/integrations/http_tracking_client.dart';
+export './src/integrations/http_tracking_client.dart' show FaroHttpOverrides;
```
aranhave, line 10: Could we keep these exports for now and deprecate them separately? They are public today, so narrowing the barrel adds another breaking API change that is not needed for trace correlation.

lib/src/tracing/faro_span_context.dart
```diff
@@ -0,0 +1,53 @@
...
+/// `pushLog`, `pushEvent`, `pushError`, or `pushMeasurement`. When a span is
+/// active, the SDK attaches the active span's context automatically, so a
+/// manual [FaroSpanContext] is only needed to override that or to correlate
+/// with a span you obtained elsewhere (for example, ids received from a backend
+/// or a WebView handoff).
+///
+/// If you already hold a [Span], use its `spanContext` getter instead of
+/// constructing this manually.
+class FaroSpanContext {
+  /// Creates a span context from a [traceId] and [spanId].
```
aranhave, line 24: I think we should document the expected OTel ID format here? `traceId` should be 32 hex characters and `spanId` 16, especially since this is the public handoff path for backend/WebView IDs.

example/lib/features/tracing/domain/tracing_service.dart
```diff
@@ -175,6 +175,47 @@ class TracingService {
     }
   }
 
+  /// Runs a span and emits all four custom telemetry types inside it.
+  ///
+  /// Used to verify automatic log-trace correlation: none of the push calls
+  /// pass a `trace`, so the SDK should stamp each signal with the active
+  /// span's `trace_id`/`span_id`.
+  Future<void> runSpanWithCustomerTelemetry(LogCallback log) async {
```
aranhave: nit: Should this be `runSpanWithCustomTelemetry`? The method and the new demo strings are describing custom telemetry.

aranhave (commented):
Looks great! 💯 
Small PR description nit: On current main, only pushLog and pushEvent have the trace parameter. Could we say those two replace trace with spanContext, while pushError and pushMeasurement newly gain spanContext?

lib/src/tracing/faro_span_context.dart
robert-northmind: thanks. added to the docs above that it should be according to the w3c spec, which is the one Otel is referring to

lib/faro.dart
robert-northmind: Good point. But maybe, from my point of view, I think it is safe to break here. :) 
This API was already supposed to be private. There was nothing in there which you should have used. It was just a mistake to expose it originally. And it sort of "maybe" relates to this change. 
Since in this change I removed some properties (the trace_id and span_id from the user-properties-map) which is exposed by one class which was exposed here. 
What do you think?

example/lib/features/tracing/domain/tracing_service.dart
robert-northmind: haha, nice. thanks. great catch :) will update it

CHANGELOG.md
```diff
@@ -7,6 +7,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  `pushMeasurement` while a span is active are now automatically stamped
+  with the active span's `trace_id`/`span_id`, matching the Faro Web SDK
+  and the OpenTelemetry logs data model. All four push methods accept an
+  optional `spanContext` parameter (new `FaroSpanContext(traceId:, spanId:)`
+  value type; use `span.spanContext` when you hold a `Span`) that always
+  takes precedence over the active span. Span context is captured at push
+  time, so buffered signals keep the span that was active when they were
+  recorded.
+
+### Changed
```
aranhave, line 23: Should we list the other breaking API changes here too? `pushLog`/`pushEvent` replace `trace` with `spanContext`, and `FaroException.trace` changes from `String` to `Map<String, dynamic>?`. The PR body covers them, but users will rely on the changelog when upgrading.

lib/faro.dart
aranhave: Yeah, that makes sense 👍 Since these were implementation details and we’re already doing a minor bump for the related breaking changes, I’m okay with removing the exports here :)

CHANGELOG.md
robert-northmind: I updated this now and documented the breaking change.
for the `FaroException.trace`, it is a breaking changes, but I am pretty sure it was not really used. since it actually didn't work before since there was a bug in the code where the trace param was not used.

It is also sort of an edge case to supply your own trace to these events. By default they will inherit the context from the current active span

aranhave (dismissed):
LGTM! 🥇

aranhave (approved):

PR #301, opened 2026-07-24
fix: forward Android native crash traces

### LINKED ISSUE #219: [BUG] Native crash stacktraces are never delivered — context key mismatch between Android SDK and Dart
## Description

On Android, native crash events (`type=crash`) always arrive in Grafana with
`context_stacktrace="No stacktrace"`, even when `ApplicationExitInfo` has
captured a trace. The root cause is a JSON key mismatch between the Android
plugin and the Dart side of the SDK, plus a missing forward of the field.

Reproduced on `faro: 0.14.0` (latest, published 2026-04-14) against a
Flutter demo app running on an Android 11 emulator (API 30). The app
triggered a Kotlin `NullPointerException` (via `value!!.length`) on the
main thread, which is a `REASON_CRASH` from `ApplicationExitInfo`.

Example event in Loki (trimmed):

```
timestamp=2026-04-17T19:53:04.255Z kind=exception level=error type=crash
  value="CRASH , status: 0"
  context_description=crash
  context_importance=100
  context_processName=com.example.flutter_mobile_o11y_demo
  context_stacktrace="No stacktrace"
  context_timestamp_readable_utc=2026-04-17T19:52:59.794Z
  sdk_name=faro-mobile-flutter sdk_version=0.14.0
```

Note the literal string `"No stacktrace"` — it is a hard-coded Dart
fallback, not a missing field.

## Root cause

There are **two compounding issues** in the SDK. Either alone would be
enough to produce the symptom.

### 1. Key mismatch between Android and Dart

The Android plugin writes the trace under the key `"trace"`:

`android/src/main/java/com/grafana/faro/ExitInfoHelper.java` (around line 159):

```java
int reasonCode = exitInfo.getReason();
if ((reasonCode == ApplicationExitInfo.REASON_CRASH ||
     reasonCode == ApplicationExitInfo.REASON_CRASH_NATIVE ||
     reasonCode == ApplicationExitInfo.REASON_ANR) &&
    exitInfo.getTraceInputStream() != null) {

    String trace = readTraceInputStream(exitInfo);
    if (trace != null && !trace.isEmpty()) {
        jsonObject.put("trace", trace);   // <-- key is "trace"
    }
}
```

The Dart side reads from `"stacktrace"` — a key that never exists:

`lib/src/faro.dart` (around line 763):

```dart
final stacktrace =
    stringifiedContext['stacktrace'] ?? 'No stacktrace';   // <-- wrong key
```

Result: the fallback literal string `'No stacktrace'` is always used.

### 2. The `"trace"` field is not forwarded in the error context

Even if the Dart code read the right key, the context passed to
`pushError` is a hard-coded allow-list of six fields and does not
include any trace:

`lib/src/faro.dart` (around line 773):

```dart
_instance.pushError(
  type: 'crash',
  value: '$reason , status: $status',
  context: {
    'description': description,
    'stacktrace': stacktrace,      // <-- must also be populated here
    'timestamp': timestamp,
    'timestamp_readable_utc': humanReadableTimestamp,
    'importance': importance,
    'processName': processName,
  },
);
```

So even after fixing the key, the trace would still be discarded unless
this map is updated.

## Expected behavior

When `ApplicationExitInfo.getTraceInputStream()` returns a non-null,
non-empty trace for `REASON_CRASH`, `REASON_CRASH_NATIVE`, or `REASON_ANR`,
that trace should appear on the resulting Faro event in a usable form
(either as a parsed set of `StackFrames` — per #212 — or at minimum as
a `context_stacktrace` text field).

## Scope

- **Affected**: All Android native crashes reported through
  `enableCrashReporter()` (JVM crashes, native NDK crashes via
  `REASON_CRASH_NATIVE`). This is the entire native-crash path on Android.
- **Not affected**: iOS crashes (separate codepath via
  `CrashReportingIntegration.swift` / PLCrashReporter) and Dart unhandled
  exceptions (via `FlutterError.onError` / zone handler).
- **Partially related**: ANR traces (#212) — those do arrive, but as raw
  text rather than structured frames. That's a separate codepath
  (`native_integration.dart` → `_getAnrStatus()`), not the
  `getCrashReport()` path discussed here.

## Suggested fix

Minimal change in `lib/src/faro.dart`:

```dart
final stacktrace =
    stringifiedContext['trace'] ?? 'No stacktrace';   // read "trace"
```

and make sure it ends up in the forwarded context (the map already has a
`'stacktrace'` key, so the fix is purely the read key).

A follow-up, aligned with #212, would be to parse the Java-format trace
into structured `StackFrames` so it renders like a Dart exception in
Grafana Frontend Observability.

## Caveat about Android itself

Independent of these SDK bugs: `ApplicationExitInfo.getTraceInputStream()`
is only documented to reliably return a trace for `REASON_ANR` and
`REASON_CRASH_NATIVE`. For `REASON_CRASH` (JVM exceptions) it is often
null on many OEMs / Android versions. So even with the SDK fixed,
JVM-crash traces may still be missing on some devices. This is an
Android limitation, not an SDK bug — but it's worth documenting.

## Environment

- `faro`: 0.14.0
- Flutter: Dart 3.11.0 (android_arm64)
- Device: Android 11 emulator (SDK 30, `sdk_gphone_arm64`)
- Repro app: [grafana/mobile-o11y-demo](https://github.com/grafana/mobile-o11y-demo) Flutter app, "Crash (simulated NPE)" button in Debug tab

## Related

- #212 — ANR stack traces are not parsed into structured Faro frames
  (related but distinct codepath; that one gets text through, this one
  drops it entirely)

aranhave:
Android sends native crash traces as `trace`, but the Dart side was looking for
`stacktrace`, so the report always fell back to `No stacktrace`.

This forwards the captured trace into the Faro exception context while keeping
compatibility with the older `stacktrace` key and the existing fallback.

Fixes #219.

### DIFF
--- a/lib/src/faro.dart
+++ b/lib/src/faro.dart
@@ -855,51 +855,56 @@ class Faro {
       if (Platform.isAndroid) {
         final crashReports = await _nativeChannel?.getCrashReport();
         if (crashReports != null) {
-          for (final crashInfo in crashReports) {
-            final crashInfoJson = json.decode(crashInfo);
-            final String reason = crashInfoJson['reason'];
-            final int status = crashInfoJson['status'];
-            // String description = crashInfoJson["description"];
-            // description/stacktrace fails to send format and sanitize before push
-
-            // Convert crashInfoJson from Map<String, dynamic> to Map<String, String>
-            final stringifiedContext = <String, String>{};
-            crashInfoJson.forEach((String key, dynamic value) {
-              stringifiedContext[key] = value?.toString() ?? '';
-            });
-
-            final description =
-                stringifiedContext['description'] ?? 'No description';
-            final stacktrace =
-                stringifiedContext['stacktrace'] ?? 'No stacktrace';
-            final timestamp = stringifiedContext['timestamp'] ?? 'No timestamp';
-            final humanReadableTimestamp = timestamp.toHumanReadableTimestamp();
-
-            final importance =
-                stringifiedContext['importance'] ?? 'No importance';
-            final processName =
-                stringifiedContext['processName'] ?? 'No processName';
-
-            _instance.pushError(
-              type: 'crash',
-              value: '$reason , status: $status',
-              fatal: true,
-              context: {
-                'description': description,
-                'stacktrace': stacktrace,
-                'timestamp': timestamp,
-                'timestamp_readable_utc': humanReadableTimestamp,
-                'importance': importance,
-                'processName': processName,
-              },
-            );
-          }
+          _reportAndroidCrashReports(crashReports);
         }
       }
     } catch (error, stacktrace) {
       log(
         'Faro: enableCrashReporter failed with error: $error',
         stackTrace: stacktrace,
+      );
+    }
+  }
+
+  @visibleForTesting
+  void reportAndroidCrashesForTesting(List<String> crashReports) {
+    _reportAndroidCrashReports(crashReports);
+  }
+
+  void _reportAndroidCrashReports(List<String> crashReports) {
+    for (final crashInfo in crashReports) {
+      final crashInfoJson = json.decode(crashInfo);
+      final String reason = crashInfoJson['reason'];
+      final int status = crashInfoJson['status'];
+
+      final stringifiedContext = <String, String>{};
+      crashInfoJson.forEach((String key, dynamic value) {
+        stringifiedContext[key] = value?.toString() ?? '';
+      });
+
+      final description = stringifiedContext['description'] ?? 'No description';
+      final stacktrace =
+          stringifiedContext['trace'] ??
+          stringifiedContext['stacktrace'] ??
+          'No stacktrace';
+      final timestamp = stringifiedContext['timestamp'] ?? 'No timestamp';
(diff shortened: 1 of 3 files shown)


lib/src/faro.dart
```diff
@@ -903,4 +865,47 @@ class Faro {
...
+      final importance = stringifiedContext['importance'] ?? 'No importance';
+      final processName = stringifiedContext['processName'] ?? 'No processName';
+
+      _instance.pushError(
+        type: 'crash',
+        value: '$reason , status: $status',
+        fatal: true,
+        context: {
+          'description': description,
+          'stacktrace': stacktrace,
```
domasx2, line 902: claude review suggests truncating to a very generous size limit. But I don't see much prior art in other SDKs, so not sure if it's a real concern

domasx2 (approved):

lib/src/faro.dart
aranhave: I checked and Android already caps these traces at roughly 1 MiB before they reach Dart, so I think we’re covered here. 👍

PR #303, opened 2026-07-27
fix(errors): report Flutter errors without stack traces

### LINKED ISSUE #271: fix(errors): report Flutter framework errors that have no stack trace instead of dropping them

aranhave:
## Description

Flutter framework errors were skipped when Flutter did not provide a stack trace. This reports them with their type and message, while continuing to include the stack when available.

Regression coverage, the error-categorization reference, and the changelog are updated.

## Related Issue(s)

Fixes #271

## Type of Change

## Checklist

## Validation

`dart tool/pre_release_check.dart` — all 6 checks passed.


benjoseph-grafana (dismissed):

robert-northmind (dismissed):
LGTM 🚀

robert-northmind (approved):

PR #304, opened 2026-07-28
feat(session): react to receiver session invalidation

### LINKED ISSUE #286: feat(session): react to server-driven session invalidation (X-Faro-Session-Status)
## Summary

The Grafana Cloud Faro receiver validates the `X-Faro-Session-Id` request header server-side. When the session has expired (inactivity or total-lifetime window), the receiver responds with the header `X-Faro-Session-Status: invalid` while still accepting the payload (HTTP 202). The Faro Web SDK reacts to this by rotating/extending its session so the client realigns with the server.

The Flutter SDK currently ignores response headers entirely, so after server-side expiry the client can keep sending a session id the server considers invalid until the client's own 15m/4h timers fire.

## Current behavior

- `FaroTransport.send` posts payloads and does not inspect the response.
- Session rotation is purely client-driven (15m inactivity / 4h lifetime), introduced in #52 / #285.

## Proposed behavior

- Inspect the response for `X-Faro-Session-Status: invalid`.
- When present, trigger a client-side rotation (new session id, `session_extend` event, `previousSession` attribute) so the client realigns with the server's view.

## Reference

- Faro Web SDK: `packages/web-sdk/src/transports/fetch/transport.ts` reads `X-Faro-Session-Status` in `handleResponse` and calls `extendFaroSession`.

## Notes

- Requires plumbing a response signal from `FaroTransport.send` back to the `SessionManager` (the transport currently has no reference to it).
- Related: #52, #283 (cross-restart persistence), #284 (re-sample on rotation).

aranhave:
## Description

The receiver can accept a payload while marking its session invalid. Flutter previously ignored that response, so it could keep using a session the receiver had already expired.

This reads `X-Faro-Session-Status: invalid` from accepted responses and rotates the matching active session. Delayed or duplicate responses for an older session are ignored.

## Related Issue(s)

Fixes #286

## Type of Change

- [x] 📝 Documentation
- [ ] 🏗️ Code refactoring
- [ ] 🧹 Chore / Housekeeping

## Checklist

Not applicable; this changes transport and session behavior without UI changes.

## Additional Notes

Validated with `dart tool/pre_release_check.dart`, the example analyzer, and a release APK build.

### DIFF
--- a/lib/src/faro.dart
+++ b/lib/src/faro.dart
@@ -232,6 +232,9 @@ class Faro {
     } else {
       Faro()._transports.addAll(config?.transports ?? []);
     }
+    for (final transport in _transports.whereType<FaroTransport>()) {
+      transport.sessionInvalidatedHandler = sessionManager.invalidateSession;
+    }
     _httpTrackingFilter.configure(
       collectorUrl: optionsConfiguration.collectorUrl,
       ignoreUrls: optionsConfiguration.ignoreUrls,
--- a/lib/src/session/session_manager.dart
+++ b/lib/src/session/session_manager.dart
@@ -10,7 +10,7 @@ enum SessionStartTrigger {
   /// The first session after the SDK started (via [SessionManager.start]).
   initial,
 
-  /// A rotation triggered by inactivity or maximum lifetime.
+  /// A rotation triggered by expiry or receiver invalidation.
   rotation,
 }
 
@@ -24,7 +24,8 @@ typedef SessionChangedListener =
       required SessionStartTrigger trigger,
     });
 
-/// Tracks session validity and rotates the session when it expires.
+/// Tracks session validity and rotates the session when it expires locally or
+/// the receiver reports it as invalid.
 ///
 /// A session expires when either:
 /// - no activity has been recorded for [inactivityTimeout]
@@ -126,6 +127,18 @@ class SessionManager {
     } else if (_activityPolicy.recordsActivity(activity)) {
       _lastActivityAt = now;
     }
+  }
+
+  /// Rotates a session that the receiver reported as invalid.
+  ///
+  /// [invalidSessionId] must still be the active session. This prevents
+  /// duplicate or delayed responses for an older session from rotating a
+  /// newer session again.
+  void invalidateSession(String invalidSessionId) {
+    if (!_isActive || _isRotating || invalidSessionId != currentSessionId) {
+      return;
+    }
+    _rotate(_currentTimeProvider());
   }
 
   /// Rotates the session and notifies listeners.
--- a/lib/src/transport/faro_transport.dart
+++ b/lib/src/transport/faro_transport.dart
@@ -27,6 +27,7 @@ class FaroTransport extends BaseTransport {
   }
 
   static const _accepted = 202;
+  static const _sessionIdHeader = 'x-faro-session-id';
   static const _sessionStatusHeader = 'x-faro-session-status';
   static const _invalidSessionStatus = 'invalid';
 
@@ -68,12 +69,10 @@ class FaroTransport extends BaseTransport {
       final headers = {
         'Content-Type': 'application/json',
         'x-api-key': apiKey,
-        'x-faro-session-id': _sessionIdResolver(),
+        _sessionIdHeader: _sessionIdResolver(),
         ...?this.headers,
       };
-      final sentSessionId = headers.entries
-          .lastWhere((entry) => entry.key.toLowerCase() == 'x-faro-session-id')
-          .value;
+      final sentSessionId = _headerValue(headers, _sessionIdHeader);
 
       final post = _httpClient?.post ?? http.post;
       final response = await _taskBuffer?.add(() {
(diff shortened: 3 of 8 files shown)


lib/src/session/session_manager.dart
```diff
@@ -10,7 +10,7 @@ enum SessionStartTrigger {
   /// The first session after the SDK started (via [SessionManager.start]).
   initial,
 
-  /// A rotation triggered by inactivity or maximum lifetime.
+  /// A rotation triggered by expiry or receiver invalidation.
```
robert-northmind, line 13: nice. better wording 👍

lib/src/transport/faro_transport.dart
```diff
@@ -54,9 +69,10 @@ class FaroTransport extends BaseTransport {
       final headers = {
         'Content-Type': 'application/json',
         'x-api-key': apiKey,
-        'x-faro-session-id': _sessionIdResolver(),
+        _sessionIdHeader: _sessionIdResolver(),
         ...?this.headers,
       };
+      final sentSessionId = _headerValue(headers, _sessionIdHeader);
```
robert-northmind, line 77: Should we add a small comment here? why we are doing this extra step and reading the sessionId out from headers?
I was confused at first here. I was like: why dont we just read out the `_sessionIdResolver()` at the top and use that.
But the reason for this, is I guess, that someone might override the sessionId with a custom one in the user provided headers?
Maybe a small comment here could help explain that?

lib/src/transport/faro_transport.dart
```diff
@@ -73,8 +89,26 @@ class FaroTransport extends BaseTransport {
           'body: ${response.body} payload:$encodedPayload',
         );
       }
+
+      if (response != null &&
+          sentSessionId != null &&
+          response.statusCode == _accepted &&
+          _headerValue(response.headers, _sessionStatusHeader) ==
+              _invalidSessionStatus) {
+        _onSessionInvalidated?.call(sentSessionId);
```
robert-northmind, line 101: Should we log here maybe when a session gets invalidated by a faroCollector?

lib/src/transport/faro_transport.dart
```diff
@@ -73,8 +89,26 @@ class FaroTransport extends BaseTransport {
...
+      }
     } catch (error) {
-      log('Error encoding payload: $error');
+      log('Error sending payload: $error');
+    }
+  }
+
+  static String? _headerValue(Map<String, String> headers, String name) {
```
robert-northmind, line 108: Should we here also add some docs, why this headers map iteration does not stop on first match? It was not obvious for me when first reading thought this.. but could also be me that is slow :D

lib/src/faro.dart
```diff
@@ -232,6 +232,9 @@ class Faro {
     } else {
       Faro()._transports.addAll(config?.transports ?? []);
     }
+    for (final transport in _transports.whereType<FaroTransport>()) {
```
robert-northmind, line 235: I think we might have a future clean up task relating to this..
I guess we have to do this since someone might pass in their own custom FaroTransport.
I think this is a bug/misstake in our API. The current only reason, that I could think of, to allow passing in custom BaseTransports, is to pass in a OfflineTransport. I think there should be no reason to create your own FaroTransport and send in.

We should maybe think or creating a new issue to clean this up. And stop exposing the FaroTransport. And offer one clean way to pass in an OfflineTransport.
Right now that API is confusing. there is the:
```dart
Faro().transports.add(
    OfflineTransport(maxCacheDuration: const Duration(days: 3)),
  );
```

and the:
```dar
FaroConfig({
    ...
    final List<FaroTransport>? transports;
    ...
);
```
The `Faro().transports` exposes the BaseTransport, and the FaroConfig exposes the FaroTransport. This is probably wrong.
We should think about which API here makes most sense.

If we could hide the FaroTransport internally, then we could just pass the `sessionInvalidatedHandler` in the constructor.


What do you think? Or am I thinking wrong. 
Not saying we change this PR. Just if we should open an issue for this.

robert-northmind (dismissed):
Nice work. looks great. Just left some minor comments and an open question on future work :)

lib/src/faro.dart
aranhave: 💯 I think you’re right. The two transport APIs are confusing, and hiding `FaroTransport` would make this wiring cleaner too. Let’s keep this PR focused and track the cleanup separately.

lib/src/transport/faro_transport.dart
aranhave: Yep, exactly. We need the ID that was actually sent, including a custom header override. I added a short comment here.

lib/src/transport/faro_transport.dart
aranhave: Yeah, that’ll help when debugging receiver responses. I added a short log without the session ID.

lib/src/transport/faro_transport.dart
aranhave: Good call. Custom headers are merged last, so the last case-insensitive match is the effective value. I added a note.

robert-northmind (approved):

PR #309, opened 2026-08-03
chore(renovate): group bumps and align npm release age with org default

domasx2:
First Renovate config for this repo (previously org defaults only).

## Summary

- `.github/renovate.json` — groups bumps by manager (`github-actions`, `npm-dependencies`, `gradle`) plus a shared `patch updates` lane, and disables the `dev.flutter.flutter-plugin-loader` marker that Renovate cannot resolve and that warns on every PR body.
- `example/webview_demo/.yarnrc.yml` — `npmMinimalAgeGate` 7d → 3d to match the org-default Renovate `minimumReleaseAge`; that mismatch is why #291 fails `renovate/artifacts`. Shortens the quarantine window set in #260.

Turns today's 11 bump PRs into ~5. Validated with `renovate-config-validator`.

_PR description authored by Claude on Domas's behalf._

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -7,6 +7,7 @@
   "separateMultipleMajor": true,
   "prHourlyLimit": 2,
   "prConcurrentLimit": 10,
+  "minimumReleaseAge": "3 days",
   "internalChecksFilter": "strict",
   "osvVulnerabilityAlerts": true,
   "vulnerabilityAlerts": {
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -31,7 +31,13 @@
       "groupName": "gradle"
     },
     {
-      "description": "Collect patch-only bumps from every manager into a single low-risk PR, matching faro-web-sdk. Placed after the manager groups so it overrides them: a Gradle or Kotlin patch rides here rather than in the gradle group. Fires only when a dependency's sole pending update is a patch.",
+      "description": "Keep the Flutter SDK pin atomic — setup-flutter fails if .fvmrc and .tool-versions disagree.",
+      "matchManagers": ["fvm", "asdf", "mise"],
+      "groupName": "flutter-sdk"
+    },
+    {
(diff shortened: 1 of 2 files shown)


.github/renovate.json
```diff
@@ -0,0 +1,51 @@
...
+      "groupName": "npm-dependencies"
+    },
+    {
+      "description": "Group Gradle/Kotlin toolchain bumps for the example app into a single PR. These are the only dependencies here with real CI coverage (build-example-app and :faro:testDebugUnitTest), so a grouped failure still points at a small and related set.",
+      "matchManagers": ["gradle", "gradle-wrapper"],
+      "groupName": "gradle"
+    },
+    {
+      "description": "Collect patch-only bumps from every manager into a single low-risk PR, matching faro-web-sdk. Placed after the manager groups so it overrides them: a Gradle or Kotlin patch rides here rather than in the gradle group. Fires only when a dependency's sole pending update is a patch.",
+      "matchUpdateTypes": ["patch"],
```
aranhave, line 41: I think this patch group is a little too broad. It also pulls pub and Swift package updates into the same PR, not just the managers grouped above.

example/webview_demo/.yarnrc.yml
```diff
@@ -6,4 +6,4 @@ enableScripts: false
 
 nodeLinker: node-modules
 
-npmMinimalAgeGate: 7d
+npmMinimalAgeGate: 3d
```
aranhave, line 9: I’d keep the 7-day gate from #260 and move Renovate to 7 days as well. That fixes the mismatch without shortening the supply-chain cooldown.

domasx2: 3 days is Grafana org default, hence why I wanted to align. Strong feelings for keeping 7 here? I think it was introduced arbitrarily

.github/renovate.json
domasx2: Robot told me that pub/swift don't generate patch updates due to how they are configured right now. But updated to exclude anyway

Also added separate group for flutter sdk, did not realize before it was a thing :-)

robert-northmind (approved):

PR #312, opened 2026-08-06
fix: preserve view metadata for unnamed routes

### LINKED ISSUE #305: bug(navigation): setViewMeta(name: null) clears known-good view meta, blanking view_name on subsequent signals
## Problem

`FaroNavigationObserver` forwards `RouteSettings.name` into `Faro().setViewMeta()` unconditionally, and `setViewMeta` overwrites the stored view meta with whatever it is handed — including `null`. A single navigation involving a route with no name therefore **erases a known-good view name**, and every signal emitted afterwards is sent with no `view.name` until the next named navigation happens.

`lib/src/faro_navigation_observer.dart` (0.17.0-beta.1), all three callbacks:

```dart
Faro().setViewMeta(name: route.settings.name);        // didPush     — name may be null
Faro().setViewMeta(name: previousRoute?.settings.name); // didPop    — may be null
Faro().setViewMeta(name: newRoute?.settings.name);     // didReplace — may be null
```

`lib/src/faro.dart:402`:

```dart
void setViewMeta({String? name}) {
  final viewMeta = ViewMeta(name);   // null -> view.name = null
  _instance.meta = Meta.fromJson({..._instance.meta.toJson(), 'view': viewMeta.toJson()});
  _instance._batchTransport?.updatePayloadMeta(_instance.meta);
}
```

`RouteSettings.name` is nullable and commonly null in real apps — dialogs, bottom sheets, modal routes, `showDialog`/`showModalBottomSheet`, and framework/router-internal wrapper routes. In those cases the SDK does not merely fail to record the new view; it **discards the view it already knew about**.

The same code path also emits `view_changed` events whose `fromView`/`toView` attributes are both null, which are not useful to consumers.

## Impact

Observed in a real go_router app (a QuickPizza demo client) over a 7-day window. `view_name` was absent from a large share of all signals:

| kind | signals | missing `view_name` |
|---|---|---|
| `event` | 512 | 225 |
| `log` | 216 | 77 |
| `measurement` | 374 | 82 |

Of 82 `view_changed` events, 36 had neither `fromView` nor `toView`.

This is silent data loss: the view dimension simply goes blank for a subset of telemetry, with nothing in the payload indicating why. Anyone grouping errors, logs, or performance by view sees an unexplained "unknown" bucket, and per-view analysis under-counts.

## Expected behaviour

A navigation to an unnamed route should never destroy the view context the SDK already has. Two established options, either of which resolves this:

1. **Ignore the null.** Guard the call so a null name is a no-op and the previous view meta is left untouched.
2. **Sticky fallback.** Keep the last known route name in the observer and fall back to it when the incoming name is null, so the ambient "current view" is never empty.

Option 2 is probably the better fit for this SDK, since `view.name` is ambient meta stamped onto every subsequent signal rather than an explicit start/end pair — there is no natural point at which "no view" is the correct state.

Additionally, `view_changed` should probably not be emitted at all when both `fromView` and `toView` resolve to null, since it carries no information.

Both patterns are common in Flutter navigation instrumentation generally, so following one of them would also match developer expectations coming from other RUM SDKs.

## Related

- Consider exposing a caller-supplied route-name extractor (rewrite a route's name, or return null to opt a route out of tracking). This is a common companion feature and would let apps name otherwise-nameless routes without reaching for `setViewMeta` directly.
- Nested-navigator support is filed separately — see the companion issue. Here #306

## Acceptance criteria

- [ ] A navigation to a route with a null `RouteSettings.name` does not clear an existing `view.name`.
- [ ] Signals emitted after such a navigation still carry the previously-known view name.
- [ ] `view_changed` is not emitted when both `fromView` and `toView` are null.
- [ ] Tests cover: push of an unnamed route, pop back to a named route, and replace with an unnamed route.

## Environment

- `faro` 0.17.0-beta.1 (behaviour is unchanged in the current `lib/src/faro_navigation_observer.dart`)
- `go_router` 17.3.0, Flutter/Dart 3.12.x
- Reproducible on both Android and iOS

iamhaseebn:
## Description

Unnamed Flutter routes currently replace the last known `view.name` with `null`, which leaves subsequent telemetry without useful view metadata. This change updates view metadata only when the destination route has a name and suppresses `view_changed` events when both route names are null. Navigation activity signals continue to be emitted for every push, pop, and replacement.

## Related Issue(s)

Fixes #305

## Type of Change

- [ ] 📝 Documentation
- [ ] 🏗️ Code refactoring
- [ ] 🧹 Chore / Housekeeping

## Checklist

- [ ] I have made corresponding changes to the documentation (not applicable)

## Verification

- `flutter test test/src/faro_navigation_observer_test.dart` (5 tests)
- `flutter analyze`
- `flutter test` (753 tests)
- `cd example/android && gradle 8.14.5 :faro:testDebugUnitTest`
- `dart tool/pre_release_check.dart` (all 6 checks)

Not applicable.

## Additional Notes

The regression coverage includes unnamed pushes, pops, replacements, returning to a named route, and the both-names-null case.

### DIFF
--- a/lib/src/faro_navigation_observer.dart
+++ b/lib/src/faro_navigation_observer.dart
@@ -1,5 +1,7 @@
 import 'package:faro/src/core/pod.dart';
 import 'package:faro/src/faro.dart';
+import 'package:faro/src/session/session_activity_kind.dart';
+import 'package:faro/src/session/session_manager.dart';
 import 'package:faro/src/user_actions/user_action_lifecycle_signal_channel.dart';
 import 'package:flutter/widgets.dart';
 
@@ -9,14 +11,18 @@ class FaroNavigationObserver extends RouteObserver<PageRoute<dynamic>> {
       lifecycleSignalChannel: pod.resolve(
         userActionLifecycleSignalChannelProvider,
       ),
+      sessionManager: pod.resolve(sessionManagerProvider),
     );
   }
 
   FaroNavigationObserver._({
     required UserActionLifecycleSignalChannel lifecycleSignalChannel,
-  }) : _lifecycleSignalChannel = lifecycleSignalChannel;
+    required SessionManager sessionManager,
+  }) : _lifecycleSignalChannel = lifecycleSignalChannel,
+       _sessionManager = sessionManager;
 
   final UserActionLifecycleSignalChannel _lifecycleSignalChannel;
+  final SessionManager _sessionManager;
 
   @override
   void didPop(Route<dynamic> route, Route<dynamic>? previousRoute) {
@@ -61,6 +67,8 @@ class FaroNavigationObserver extends RouteObserver<PageRoute<dynamic>> {
         'view_changed',
         attributes: {'fromView': fromView, 'toView': toView},
       );
+    } else {
+      _sessionManager.checkSession(activity: SessionActivityKind.active);
     }
     _lifecycleSignalChannel.emitActivity(source: activitySource);
   }
--- a/lib/src/faro_navigation_observer.dart
+++ b/lib/src/faro_navigation_observer.dart
@@ -21,42 +21,47 @@ class FaroNavigationObserver extends RouteObserver<PageRoute<dynamic>> {
   @override
   void didPop(Route<dynamic> route, Route<dynamic>? previousRoute) {
     super.didPop(route, previousRoute);
-    Faro().setViewMeta(name: previousRoute?.settings.name);
-    Faro().pushEvent(
-      'view_changed',
-      attributes: {
-        'fromView': route.settings.name,
-        'toView': previousRoute?.settings.name,
-      },
+    _recordNavigation(
+      fromView: route.settings.name,
+      toView: previousRoute?.settings.name,
+      activitySource: 'navigation.pop',
     );
-    _lifecycleSignalChannel.emitActivity(source: 'navigation.pop');
   }
 
   @override
   void didPush(Route<dynamic> route, Route<dynamic>? previousRoute) {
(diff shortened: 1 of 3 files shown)


lib/src/faro_navigation_observer.dart
```diff
@@ -21,42 +21,47 @@ class FaroNavigationObserver extends RouteObserver<PageRoute<dynamic>> {
...
+
+  void _recordNavigation({
+    required String? fromView,
+    required String? toView,
+    required String activitySource,
+  }) {
+    if (toView != null) {
+      Faro().setViewMeta(name: toView);
+    }
+    if (fromView != null || toView != null) {
```
aranhave, line 65: An unnamed-to-unnamed navigation now skips `pushEvent`, which is the only call here that reaches `SessionManager.checkSession`. The lifecycle channel only extends the active user action. Do we still want this navigation to refresh session inactivity while suppressing the empty event?

aranhave (commented):
Looks good overall 🙂 I left one question.

lib/src/faro_navigation_observer.dart
iamhaseebn: Good catch - yes, navigation should still refresh session inactivity even when both route names are null. I updated the no-event path to call SessionManager.checkSession with active activity directly, while still suppressing the empty view_changed event, and added regression coverage for both behaviors. The focused navigation tests, full 753-test Flutter suite, analyzer, and formatting all pass.

aranhave (approved):
LGTM, thanks for addressing the session refresh case 🙂

PR #313, opened 2026-08-06
fix(app-start): stop measuring process age as cold start

### LINKED ISSUE #302: [BUG] Cold start duration measures process age, not user-visible startup
## Summary

`appStartDuration` for cold starts is computed as "time since the process started", not "time until the app became usable". On Android the operating system routinely starts the app process without any user involvement, so the reported cold start is the age of the process at the moment Faro asks for it — frequently minutes, hours, or days.

## Evidence

A production Flutter app (Faro Flutter SDK 0.16.0), 6,070 cold-start samples over 2 days:

| percentile | value |
| --- | --- |
| p10 | 348 ms |
| p50 | 1.2 s |
| p75 | 38.8 s |
| p90 | 3.3 hours |
| p99 | 50.7 hours |
| max | 13.9 days |

Split by platform:

| OS | samples | p50 | p75 | > 60 s |
| --- | --- | --- | --- | --- |
| Android | 2,470 | 102 s | ~2 h | 51.6% |
| iOS | 3,589 | 816 ms | 1.34 s | 5.9% |

More than half of Android cold starts report over a minute, and 15% of all samples exceed one hour. The distribution is bimodal: ~72% below 10 s, ~3% between 10 and 60 s, ~24.5% above 60 s. There is no plausible startup workload in the gap, which is what you would expect if the large values are a different phenomenon rather than slow launches.

## Root cause

`android/src/main/java/com/grafana/faro/FaroPlugin.java`:

```java
private long getAppStart(){
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
        return SystemClock.elapsedRealtime() - Process.getStartElapsedRealtime();
    }
    return 0;
}
```

`Process.getStartElapsedRealtime()` is when the *process* started. Android starts app processes for reasons that never show UI — FCM messages, WorkManager/JobScheduler jobs, broadcast receivers, ContentProvider initialization. When the user later opens the app, the elapsed time since process start is reported as the cold start.

`ios/faro/Sources/faro/AppStart.swift` has the same shape via `sysctl` on `kp_proc.p_starttime`. iOS is less affected, but iOS 15+ prewarming pre-launches the process in exactly the same way, which likely accounts for the 5.9% tail above 60 s.

Separately, on both platforms the interval *ends* when Dart calls `getAppStart()`, not when the first frame is rendered. So even the well-behaved values measure "time to Faro initialization" rather than time to initial display.

## Impact

Any consumer computing a percentile over this field gets a meaningless number. Grafana Frontend Observability's mobile overview showed a p75 cold start of 39.6 s for this app. We have added a client-side workaround that discards measurements above 60 s, which brings the displayed p75 to 1.48 s, but this throws away half of all Android samples and cannot distinguish a genuinely slow launch from a background process start.

## Suggested fix

Only report a cold start when the launch was user-visible, and end the interval at first frame:

**Android**
- API 35+: use [`ApplicationStartInfo`](https://developer.android.com/reference/android/app/ApplicationStartInfo), which reports the start reason, the start type, and timestamps for each launch phase. Discard starts whose reason is not user-initiated.
- Below API 35: require that an Activity was actually created, and measure to first frame (`reportFullyDrawn()`, or `ViewTreeObserver.OnDrawListener` / `Choreographer`). Discard when no Activity appears within a short window of process start.

**iOS**
- Skip prewarmed launches: `ProcessInfo.processInfo.environment["ActivePrewarm"] == "1"`.
- Measure from `didFinishLaunching` to first frame rather than from process start to SDK initialization.

**Both**
- Consider a sanity bound in the SDK so obviously invalid values are never emitted.

## References

- Android app startup: https://developer.android.com/topic/performance/vitals/launch-time
- `ApplicationStartInfo`: https://developer.android.com/reference/android/app/ApplicationStartInfo
- iOS prewarming: https://developer.apple.com/documentation/uikit/app_and_environment/responding_to_the_launch_of_your_app/about_the_app_launch_sequence

robert-northmind:
## Description

Cold start duration measured how long the process had existed, not how long a
user waited. Processes the system starts in the background — push messages,
jobs and broadcasts on Android, prewarming on iOS — reported the age of the
process as the cold start, reaching hours in production data.

Android now reports a cold start only for launches it can show were
user-visible; iOS cannot tell a background launch apart, so it instead
measures a prewarmed launch from when the SDK loaded. Both platforms discard
anything over 60 seconds. **Expect cold start volume to drop on Android.**

Also fixes an unrelated, pre-existing bug found while validating: every launch
emitted a phantom warm start of a few milliseconds alongside its cold start.
Kept as a separate commit.

See `doc/Reference.md` for per-platform detection and known limitations.

## Related Issue(s)

Fixes #302

## Type of Change

## Checklist

## Additional Notes

Validated end to end against Grafana Cloud on a physical iPhone 15 Pro
(iOS 26.5), a physical Android 16 device, and emulators on Android 11, 15 and
16. Each launch now produces exactly one cold start and no phantom warm start,
and genuine background returns still report a warm start.

The iOS prewarm path is the gap in that validation. Every launch captured had
`prewarmed=0`: iOS decides when to prewarm on its own, and forcing
`ActivePrewarm` on a simulator produces a process that stays suspended and
never renders a frame. That arm is covered by unit tests but has not been seen
on a device.

`ApplicationStartInfo` (API 35+) was implemented and then deliberately removed.
Its records persist across reboots and the record for the launch being measured
is not filed by the time the first frame renders, so it misclassified real
launches — it replaced a correct answer with a wrong one. The commit body has
the detail.

This PR also adds the first iOS native test coverage the repo has had, along
with a macOS CI job that gives us our only iOS compile coverage.


lib/src/faro.dart
```diff
@@ -356,6 +354,19 @@ class Faro {
...
+  /// Using that frame rather than the next one after `init` keeps the
+  /// measurement from stretching when a host app initialises Faro late. The
+  /// future completes immediately when the frame has already been rendered.
+  void _reportColdStartAfterFirstFrame() {
+    unawaited(
+      WidgetsBinding.instance.waitUntilFirstFrameRasterized.then(
+        (_) => _nativeIntegration.getAppStart(),
+      ),
+    );
+  }
```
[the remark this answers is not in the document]
robert-northmind: Confirmed, and fixed in 7366a96 — thanks, this was a real inaccuracy.

Both native implementations compute the duration at the moment they are
called (`SystemClock.uptimeMillis() - Process.getStartUptimeMillis()` on
Android, `wallclockMillis() - processStart` on iOS), so
`waitUntilFirstFrameRasterized` only gates *when* we ask. An app that calls
`Faro.init` after the first frame finds the future already complete and is
measured to `init`.

It is still an improvement rather than a regression: the previous
`addPostFrameCallback` fired on the frame *after* init, which is always at
least as late, and on an idle screen can be much later since Flutter
produces no frames when nothing changes. What was wrong was the wording, so
the helper comment, the changelog entry and `doc/Reference.md` now state
that the interval ends at the first rasterized frame *or* at `Faro.init`
when the app initialises after that frame.

Ending it exactly at the frame would mean taking the endpoint from
`FrameTiming.rasterFinishWallTime`, as the Sentry Flutter SDK does. That
does not fix late init either, and it would require computing the duration
in Dart against a wall-clock process start, giving up the monotonic clock
we chose on Android specifically to be immune to clock changes. Tracking
that separately rather than in this PR.

doc/Reference.md
```diff
@@ -172,7 +172,80 @@ Faro().runApp(
...
+#### Which cold starts are reported
+
+Both platforms start app processes without any user involvement. Android forks
+a process to deliver a push message, run a scheduled job or handle a broadcast,
+and that process then stays in the LRU cache indefinitely. iOS *prewarms* apps,
+creating the process and loading its libraries ahead of the user tapping the
+icon. In both cases the time since process start can be hours, and reporting it
+as a cold start makes any percentile over the metric meaningless.
+
+The SDK therefore only reports a cold start when it can establish that the
```
aranhave: iOS always marks the launch as user-visible when it has a valid duration and the Dart code notes that background launches cannot be detected there. Can we narrow this claim here and in the changelog?

example/ios/RunnerTests/AppStartTrackerTests.swift
```diff
@@ -0,0 +1,45 @@
...
+  func truncatesPlausibleDuration() {
+    #expect(AppStartTracker.millisAsInt(1234.9) == 1234)
+  }
+
+  @Test("carries the keys the Dart side reads")
+  func payloadUsesTheAgreedKeys() {
+    let metrics = AppStartTracker.coldStartMetrics()
+
+    #expect(metrics["isUserVisibleColdStart"] as? Bool == true)
+    #expect(metrics["prewarmed"] as? Bool == false)
```
aranhave, line 97: Did the iPhone validation actually capture `prewarmed=1`? This test only exercises the default `false` path, so I can’t see the prewarm branch covered yet.

lib/src/integrations/native_integration.dart
```diff
@@ -88,16 +88,54 @@ class NativeIntegration implements Disposable {
     }
   }
 
+  /// Upper bound on a plausible user-visible cold start.
+  ///
+  /// The native side already rejects process starts it can prove were not
+  /// user-initiated, but that proof is only authoritative on Android 15+, and
```
aranhave: nit: Is the Android 15+ wording left over from the old `ApplicationStartInfo` path? The provider sample now runs on every supported Android version.

aranhave (commented):
Looks good overall 🙂 I left two questions and one small nit.

lib/src/integrations/native_integration.dart
robert-northmind: Confirmed, and fixed in 6c69a84 — it is leftover from `ApplicationStartInfo`.
The provider sample is now the only mechanism, on every supported version, so
the caveat no longer describes anything. Reworded to say what actually
constrains the bound: Android can show a launch was user-initiated, iOS
cannot, so on iOS this is the only guard.

doc/Reference.md
robert-northmind: You are right, and fixed in 6c69a84. On iOS `isUserVisibleColdStart` is
`duration != nil` — it says a clock anchor was found, nothing about the
launch — so the flag name promises more than the platform can deliver. Only
Android decides user-visibility, from the process importance sampled at
startup. The docs generalised the Android guarantee to both platforms.

The Reference docs and the changelog now state per platform what is
established: iOS reports every launch it can measure, re-bases a prewarmed
one onto the SDK load anchor, and beyond that has only the 60-second cap.
The two `NativeIntegration` comments carried the same overstatement and were
narrowed with it.

Renaming the field to match what it means on both sides would be clearer
still, but it is a channel key change, so I would rather do that separately
than widen this PR.

example/ios/RunnerTests/AppStartTrackerTests.swift
robert-northmind: No, it did not. Every measurement the iPhone 15 Pro produced had
`prewarmed=0`, and I could not force otherwise: iOS decides when to prewarm
on its own, and `SIMCTL_CHILD_ActivePrewarm=1` on a simulator produces a
process that stays suspended and never renders a frame, so nothing is
reported at all. The prewarm arm is therefore unvalidated on a device, and I
do not have a way to validate it short of catching one in the wild.

Your second point stands too, and is fixed in d32bb5b. The arm had no
coverage and could not get any: `ActivePrewarm` is read once when the SDK
loads and iOS clears it afterwards, so a test process never reaches it
through `coldStartMetrics`. The anchor choice is now a function that takes
its inputs, as `resolveUserVisible` already does on Android, and both arms
are tested. The prewarm case feeds it a process start an hour before the SDK
loaded and asserts it measures the launch rather than the hour.

doc/Reference.md
aranhave: Thanks, the platform split is clear now. Two summary spots still need the same caveat: the PR title/description imply every cold start is known to be user-visible, which is not true on iOS and the changelog says late init ends at the first frame even though it measures to `Faro.init`.

robert-northmind: Both correct, thanks — I narrowed the docs and then left the two summaries
saying the old thing.

The PR title and description are updated: the title no longer claims a
user-visible guarantee, and the description carries the same per-platform
split as the changelog. I also noted there what the iPhone run did not cover,
since the prewarm arm never produced `prewarmed=1` on a device.

The changelog line is fixed in a66b1ca. It said the interval ends at the first
rasterized frame and then mentioned late init, which is the one case where
that is not what happens — an app initialising after that frame is measured to
`Faro.init`, because the native side measures up to the moment it is called.
`doc/Reference.md` and the helper comment already said this; the changelog was
the last place with the blanket claim.

robert-northmind:
@grafana/mobile-o11y-engineers  could i get another round of Reviews on this one. Thanks

aranhave (approved):
LGTM

PR #329, opened 2026-08-14
fix(session): attribute recovered crashes to original session

### LINKED ISSUE #151: feat(crash): map crash/ANR telemetry to crashed session instead of next session
## Summary

Map crash and ANR telemetry to the session where they occurred (previous session) instead of the session where they are detected (current session).

## Problem

When a native crash or ANR occurs, the app terminates before it can report the crash. On the next app start, the SDK detects and reports the crash, but currently attaches it to the **new session** rather than the **crashed session**.

This creates misleading telemetry:
- The new session appears to have errors when it was actually error-free
- The crashed session looks clean when it actually had a fatal error
- Users cannot see what happened before the crash in the correct session context

### Affected Platforms

- **Android**: Native crashes and ANRs detected via `ApplicationExitInfo`
- **iOS**: Native crashes detected via PLCrashReporter

### Current Behavior

```
Session A (crashed)     → No crash telemetry attached
  ↓ [app terminates]
Session B (clean start) → Crash from Session A incorrectly attached here
```

### Expected Behavior

```
Session A (crashed)     → Crash telemetry correctly attached
  ↓ [app terminates]
Session B (clean start) → Clean, no crash telemetry
```

### Android Implementation

The Android implementation uses `ApplicationExitInfo` API (Android 11+) to detect crashes and ANRs. Key files:
- `android/src/main/java/com/grafana/faro/ExitInfoHelper.java` - Retrieves crash data
- `android/src/main/java/com/grafana/faro/FaroPlugin.java` - Processes crash reports
- `lib/src/integrations/native_integration.dart` - Receives crash data from native

**Important**: The crash list from `ApplicationExitInfo` is **not cleared** when read. If the app had 3 old crashes, we would keep reading them each time the app starts. Current logic uses `SharedPreferencesService` to track handled crashes and avoid duplicates (see `ExitInfoHelper.filterHandledExitInfo()`).

### iOS Implementation

The iOS implementation uses PLCrashReporter. Key files:
- `ios/Classes/CrashReportingIntegration.swift` - Detects and processes crashes
- Crash reports are purged after reading via `crashReporter.purgePendingCrashReport()`

### Session Management

Sessions are managed by:
- `lib/src/session/session_id_provider.dart` - Generates session IDs
- Session ID is created on SDK initialization

### High-Level Approach

1. **Persist session ID** before app termination
   - Store current session ID to disk/SharedPreferences on each session start
   - Update on session rotation if applicable

2. **Retrieve previous session ID** on app start
   - Before creating new session, read the last persisted session ID
   - This represents the session that was active when the app last ran

3. **Map crashes to previous session**
   - When crash data is detected on startup, attach it to the previous session ID
   - Include `crashedSessionId` in crash telemetry metadata

4. **Handle edge cases**
   - **No previous session**: First app launch, no crashes to map
   - **Multiple crashes**: If app crashed multiple times, each crash should map to its respective session
   - **Time gaps**: User may not open app for days/weeks after crash - ensure old session data isn't discarded
   - **Duplicate detection**: Ensure existing duplicate prevention logic (Android `SharedPreferencesService`) still works with new session mapping

#### Android-Specific
- Integrate with existing `SharedPreferencesService` for session persistence
- Ensure `filterHandledExitInfo()` logic remains compatible
- Handle cases where `ApplicationExitInfo` returns multiple crashes from different sessions

#### iOS-Specific
- Persist session ID to UserDefaults or file storage
- Coordinate with existing crash purge logic

#### Cross-Platform
- Session persistence should be platform-agnostic where possible
- Consider session rotation/timeout scenarios
- Ensure crash metadata includes both `sessionId` (current) and `crashedSessionId` (previous)

## Acceptance Criteria

- [ ] Session ID is persisted to disk on each session start/rotation
- [ ] Previous session ID is retrieved on app startup before new session creation
- [ ] Crash telemetry includes `crashedSessionId` attribute with the previous session ID
- [ ] Existing duplicate crash detection logic continues to work correctly
- [ ] Solution handles edge cases: first launch, multiple crashes, time gaps
- [ ] Behavior is consistent across Android and iOS platforms
- [ ] Tests cover session persistence, crash mapping, and edge cases
- [ ] Documentation updated to explain crash-to-session mapping behavior

aranhave:
Keeps recovered Android crashes, ANRs, and iOS crashes attached to the session that crashed instead of the new session created after restart.

Fixes #151

#### Validation:

- iOS simulator native crash and relaunch: the recovered crash stayed on session `4qXTC2rax8`, while the new session `7JFtFN9cTD` linked back through `previousSession`.

https://github.com/user-attachments/assets/3bc0d251-fe38-4736-b8f6-286c4981701b

Reliable iOS crash transport remains tracked separately in #269.

### DIFF
--- a/lib/src/faro.dart
+++ b/lib/src/faro.dart
@@ -214,7 +214,7 @@ class Faro {
       installationId: '$installationId',
     );
 
-    final persistedPreviousSessionId = await _initializeSessionPersistence(
+    final persistedPreviousSession = await _initializeSessionPersistence(
       optionsConfiguration,
     );
 
@@ -265,6 +265,7 @@ class Faro {
         app: _instance.meta.app!,
         apiKey: optionsConfiguration.apiKey,
         collectorUrl: optionsConfiguration.collectorUrl ?? '',
+        recoveredSession: persistedPreviousSession,
       );
     }
     if (Platform.isAndroid || Platform.isIOS) {
@@ -278,7 +279,9 @@ class Faro {
     }
     await FaroOtelBootstrap.initialize();
     // Announce the initial session; the listener emits `session_start`.
-    sessionManager.start(previousSessionId: persistedPreviousSessionId);
+    sessionManager.start(
+      previousSessionId: persistedPreviousSession?.currentSessionId,
+    );
     await _sessionPersistence?.flush();
     _reportColdStartAfterFirstFrame();
 
@@ -387,7 +390,9 @@ class Faro {
     );
   }
 
-  Future<String?> _initializeSessionPersistence(FaroConfig options) async {
+  Future<PersistedSessionRecord?> _initializeSessionPersistence(
+    FaroConfig options,
+  ) async {
     if (!_isMobilePlatform()) {
       return null;
     }
@@ -427,7 +432,7 @@ class Faro {
       }
 
       _sessionPersistence = persistence;
-      return (await persistence.load())?.currentSessionId;
+      return persistence.load();
     } catch (error) {
       log('Faro: Session persistence unavailable: $error');
       _sessionPersistence = null;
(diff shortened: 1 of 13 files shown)


benjoseph-grafana (dismissed):

lib/src/faro.dart
```diff
@@ -1042,19 +1066,39 @@ class Faro {
...
+      final recoveredCrashMeta = recoveredSession == null
+          ? null
+          : _metaForRecoveredSession(recoveredSession);
+      final metadata = (recoveredCrashMeta ?? meta).toFaroJson();
       metadata['app'] = app.toJson();
       metadata['apiKey'] = apiKey;
       metadata['collectorUrl'] = collectorUrl;
-      if (Platform.isIOS) {
+      metadata['reportPendingCrash'] =
+          (recoveredSession?.isSampled ?? true) && enableDataCollection;
```
benjoseph-grafana: Android drops the crash if recovered session is null (https://github.com/grafana/faro-flutter-sdk/pull/329/changes#diff-4999f2cb1c837c711a1e04172c8afdfad17dea5fc402200448507145c1a3d0eeR1172).
Can we have the same behaviour on both platforms?

aranhave: I checked this path. Android only drops the crash when persistence is active but no persisted session matches. Treating an empty history as the fallback would attach the old crash to the new session, which is what #151 is trying to prevent. I think we should keep that behavior and make iOS skip it in the same case. Does that match what you had in mind?

benjoseph-grafana: yes.

benjoseph-grafana (approved):

PR #337, opened 2026-08-18
chore(webview): update vulnerable dependencies

aranhave:
## Description

Refreshes the webview demo lockfile to use patched PostCSS, Nano ID, and brace-expansion releases before the vulnerability SLO expires. These are example-only dependencies and do not affect the SDK runtime.

## Type of Change

- [x] 🧹 Chore / Housekeeping


yduartep (approved):
LGTM

Issue #340, opened 2026-08-26
test(session): verify Android secondary-engine ownership on device


aranhave:
## Summary

Add a small Android device harness that starts an activity-backed Flutter engine and a secondary/headless engine in the same process, then verifies the supported session ownership model.

## Supported behavior

One root engine per process may own persisted session state. Other engines use independent in-memory sessions. Android headless engines report `dart_isolate_name=headless`; activity-backed engines report `main`.

## Acceptance criteria

- [x] The harness runs the foreground and headless engines concurrently on an Android emulator or device.
- [x] Only one engine owns persisted session state at a time; the other falls back to an in-memory session without changing the owner's persisted chain.
- [x] Runtime metadata distinguishes `main` and `headless`.
- [x] Releasing the owner allows a newly created engine to claim persistence.
- [x] Telemetry shows no duplicate or corrupted session links and no cross-engine crash attribution.
- [x] Commands and observed results are documented, and the harness is retained as a repeatable test target or its reusable setup is captured in test code or documentation.

## References

- Runtime investigation: #6
- Headless classification: #333 and #334
- Shared session epic: grafana/frontend-o11y-knowledge-workbench#57

aranhave:
Verified from merged `main` at [`81ff530`](https://github.com/grafana/faro-flutter-sdk/commit/81ff53090d770b4d24aa23c423dcc254f031fefe) on the `quickpizza_pixel_35` Android 15 / API 35 ARM64 emulator.

```sh
cd example
flutter pub get
flutter build apk
cd android
./gradlew :app:connectedDebugAndroidTest \
  -Pandroid.testInstrumentationRunnerArguments.class=com.grafana.faro_example.SecondaryEngineSessionTest
```

Both device tests passed. The run covered both startup orders and confirmed one persistence owner at a time, `main`/`headless` metadata, one telemetry probe from each engine, isolation of the secondary engine's reset, handoff to a replacement engine, an intact durable session chain, and one recovered crash attributed to the original owner session without cross-engine or duplicate attribution.

PR #341, opened 2026-08-26
fix(session): scope Android crash recovery to the persistence owner

aranhave:
## Description

Makes recovered Android crashes follow the same process-scoped owner as durable session persistence. A root Flutter engine that does not own persistence now skips historical crash recovery instead of consuming a crash with an unrelated live session. If ownership is unavailable, recovery is left for a later launch.

Adds a retained Android device harness that runs Activity-backed and headless engines in both startup orders. It verifies first-come ownership, non-owner isolation, ownership handoff, runtime metadata, durable session-chain integrity, and recovered-crash attribution.

## Related Issue(s)

Related to #340

## Type of Change

- [x] 📝 Documentation
- [x] 🧹 Chore / Housekeeping

## Checklist

Not applicable; there is no UI change.

## Additional Notes

Crash recovery delivery remains asynchronous, so the owning engine must stay alive until its configured transport accepts the payload. The documentation covers this requirement and the fail-closed behavior when runtime ownership cannot be established.

After merge, the retained device test will be rerun from `main` before #340 is closed.

### DIFF
--- a/.github/workflows/flutter_checks.yml
+++ b/.github/workflows/flutter_checks.yml
@@ -77,6 +77,10 @@ jobs:
         working-directory: example/android
         run: ./gradlew :faro:testDebugUnitTest
 
+      - name: Compile Android instrumentation tests
+        working-directory: example/android
+        run: ./gradlew :app:assembleDebugAndroidTest
+
   test-ios:
     name: build & test iOS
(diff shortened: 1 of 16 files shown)


robert-northmind (approved):
LGTM

PR #342, opened 2026-08-27
feat(webview): propagate an app-owned span

robert-northmind:
## Description

`FaroWebViewBridge.instrumentedUrl` always started its own span, so the
`traceparent` it injected into the WebView URL could never match a span the
app already used to track that WebView. Passing the same `spanName` did not
connect them, the app's span and the web app's telemetry landed in two
unrelated traces. And because every call started a fresh span, the
`traceparent` changed on each call, which breaks WebViews that load more than
one URL during their lifetime.

`instrumentedUrl` now accepts an optional `span` and propagates that span's
trace context. Span lifecycle follows ownership: the bridge only ends spans it
created. A span you pass in is never given attributes or a status and is never
ended, so repeated calls keep injecting the same `traceparent`, and nothing the
bridge does can truncate a span that has to outlive a single load. The default
path is unchanged.

## Related Issue(s)

None.

## Type of Change

- [ ] 💥 Breaking change
- [x] 📝 Documentation
- [ ] 🏗️ Code refactoring
- [ ] 🧹 Chore / Housekeeping

## Checklist

## Validation

`dart tool/pre_release_check.dart`: all 6 checks pass.

Verified end to end by running the example app on an iOS simulator through both
WebView entry points, then inspecting the emitted telemetry in Mobile o11y with
gcx. The app-owned span keeps a single trace across three `instrumentedUrl`
calls, with the React app's request parented to it, while the default path
still creates a span per load. Session correlation (`session.linked` and the
`session.parent_*` attributes on the web session) is unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   takes precedence over the active span. Span context is captured at push
   time, so buffered signals keep the span that was active when they were
   recorded.
+- **App-owned spans in `FaroWebViewBridge`.** `instrumentedUrl` accepts an
+  optional `span` and propagates that span's `traceparent` instead of
+  creating a span of its own, so repeated calls during one WebView session
+  keep injecting the same `traceparent`. The bridge leaves a span you pass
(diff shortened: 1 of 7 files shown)


domasx2 (approved):
LGTM!

PR #344, opened 2026-08-28
chore: bump version to v0.17.0-beta.3

robert-northmind:
## Description

Prepares the `0.17.0-beta.3` release. Bumps the version in `pubspec.yaml`,
`ios/faro.podspec`, `android/build.gradle`, `lib/src/util/constants.dart` and
`example/pubspec.lock`, and moves the accumulated `CHANGELOG.md` Unreleased
content under a dated `0.17.0-beta.3` heading.

Release contents: 5 Added, 7 Changed (4 marked BREAKING), 1 Deprecated,
12 Fixed. Staying on the beta line because the 1.1.0 OTel dependency is still
a prerelease.

## Validation

Verified end-to-end against Grafana Cloud, not just unit tests, by pointing the
QuickPizza Flutter demo app at this branch (iOS simulator, dev stack):

- Compiles and runs unchanged despite the four breaking changes.
- Cold start reports a plausible 3.3s with the new `prewarmed` field, and no
  phantom warm start alongside it.
- Log-trace correlation resolves every correlated signal to a real Tempo span
  (5/5 traces, 0 dangling refs).
- The 15-minute inactivity rotation fires correctly — periodic vitals no longer
  keep an idle session alive.
- Session linking holds across both rotation and repeated cold starts, emitting
  `session_start` with `previousSession` and no `session_extend`.

## Related Issue(s)

None — release chore.

## Type of Change

- [x] 🧹 Chore / Housekeeping

## Checklist

- [x] I have made corresponding changes to the documentation — n/a, no API change
- [x] I have added tests that prove my fix is effective — n/a, no behaviour change
- [x] I have updated the CHANGELOG.md — inverted for a release PR: this moves
      the existing Unreleased content under the new version heading

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.17.0-beta.3] - 2026-08-27
+
 ### Added
 
 - Add `Faro.resetSession()` for logout, account changes, and custom session
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,5 +1,5 @@
 group 'com.grafana.faro'
-version '0.17.0-beta.2'
+version '0.17.0-beta.3'
 
 buildscript {
     repositories {
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -199,7 +199,7 @@ packages:
       path: ".."
       relative: true
     source: path
-    version: "0.17.0-beta.2"
+    version: "0.17.0-beta.3"
   ffi:
     dependency: transitive
     description:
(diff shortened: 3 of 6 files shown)


domasx2 (approved):
why beta-3 and not 0.17.1?

robert-northmind:
> why beta-3 and not 0.17.1?

Good question. 
We sort of need to keep the `beta.x` pattern since we rely on the OTel Dart which is still in pre-release.
pub.dev won't let us publish a stable package that depends on a pre-release package.
We will get an error message like this:
`Packages dependent on a
pre-release of another package should themselves be published as a pre-release
version`

The OTel Dart will soon be out of pre-release, that is the thing we are working on separately in the process to get OTel Dart accepted into Official OTel.

PR #349, opened 2026-09-11
feat(http): adopt stable span attributes and redact HTTP URLs

### LINKED ISSUE #346: feat(http-tracing): adopt stable span attributes and preserve Faro events
## Goal

Adopt stable HTTP client span attributes while preserving the existing
`faro.tracing.fetch` custom event contract. Keep the event name, fields, string serialization, duration and trace/session
correlation unchanged. Apply the same URL credential and sensitive-query redaction
to span `url.full` and event `http.url`; other event values remain unchanged.

The Faro event is an independent telemetry contract. Capture its existing HTTP
metadata separately from OTel span attributes so span schema changes do not
implicitly change the event payload. Existing event metadata is not a set of
temporary legacy span aliases.

## Core span mapping

| Span field | Expected value/type |
| --- | --- |
| http.request.method | Request method, string; existing known-method normalization |
| url.full | Absolute request URL with required sanitization, string |
| server.address | Host without port, string |
| server.port | Effective port including scheme default, integer |
| http.response.status_code | Actual response code, integer; absent without a response |

Set request fields at span start. Do not invent a stable response code of 0.
Preserve the status and naming behavior delivered by #24 and #109. Spans without
a response use exception `error.type`; legacy events retain their previous
status-0 behavior. Body failures keep the actual response status.

## Event compatibility

- Keep existing event HTTP URL, host, scheme, method, status, user agent,
  Content-Length sizes and content type, including their previous values and
  availability, except for URL redaction. The known-method fields added by #109 remain present.
- Redact URL credentials and the documented sensitive query keys in both spans
  and events. Keep the actual request URL unchanged.
- Preserve session/action correlation, trace/span IDs and duration.
- Do not add event-only metadata to OTel spans or rename Content-Length values
  to total request/response wire sizes.
- Keep event storage and query consumers compatible with the existing schema and redacted URL values.

## Acceptance

- [ ] Core span attributes have correct values and serialized types, including
  effective default ports and response-code absence when no response exists.
- [ ] Automatic HTTP events use `faro.tracing.fetch` and retain the existing
  event fields and supported serialization, with the approved URL redaction.
- [ ] Tests cover the independent contracts, metadata isolation, duration,
  session/action and trace correlation, and response/error behavior.
- [ ] Example and demo success, 4xx and 5xx requests verify exported versus stored
  span/event data with both apps on Android and iOS.
- [ ] Targeted manual HTTP event-view verification confirms existing queries
  still show the requests, errors, duration and domain grouping.
- [ ] Documentation distinguishes changed span fields from retained event
  fields, including URL and error behavior.

No additional-client coverage, dynamic/custom templates, expanded HTTP lifecycle
or failure scenarios, or broad product UI checks are added by this work item.

robert-northmind:
## Description

Automatic HTTP client spans use stable attributes: `http.request.method`,
`url.full`, `server.address`, `server.port`, and `http.response.status_code`.
Failures without a response omit the span status code and record an exception
`error.type`.

The `faro.tracing.fetch` event keeps its existing fields and string serialization.
Its HTTP metadata is captured separately from span attributes, preserving status,
size, duration and trace/session/action correlation. Span `url.full`, event `http.url`, and fallback `network_error` log URLs
redact URL credentials and the documented sensitive query keys, including common
application credentials such as `token`, `api_key`, and `password`.
The request URL is unchanged; exact telemetry URL filters use redacted values.
The public Dart API and outer Faro payload are unchanged.

## Related Issue(s)

Fixes #346

## Type of Change

- [x] 💥 Breaking change (HTTP span attributes and sensitive URL values)
- [x] 📝 Documentation

## Checklist

## Additional Notes

All six local pre-release checks passed. Initial example and QuickPizza runs
covered Android and iOS. Fresh Android runs of both apps verified eight matching
span/event pairs with the expanded URL redaction against stored telemetry. A
truncated response also verified the redacted fallback log and trace correlation.
The existing HTTP event query still returns the requests. Manual HTTP-view
verification of this final revision is pending.

Public redaction configuration is planned as a separate follow-up.

### DIFF
--- a/lib/src/integrations/http_url_redaction.dart
+++ b/lib/src/integrations/http_url_redaction.dart
@@ -1,0 +1,38 @@
+/// Redacts the HTTP span URL without changing the request or Faro event URL.
+///
+/// OTel HTTP client URL redaction rules (the query-key list is Development):
+/// https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span
+/// Preserves the encoding/order of non-sensitive query parameters.
+String redactHttpSpanUrl(Uri url) {
+  const sensitiveKeys = {
+    'X-Amz-Signature',
+    'X-Amz-Credential',
+    'X-Amz-Security-Token',
+    'AWSAccessKeyId',
+    'Signature',
+    'sig',
+    'X-Goog-Signature',
+  };
+  final query = url.query
+      .split('&')
+      .map((part) {
+        final separator = part.indexOf('=');
+        final key = separator < 0 ? part : part.substring(0, separator);
+        try {
+          if (sensitiveKeys.contains(Uri.decodeQueryComponent(key))) {
+            return '$key=REDACTED';
+          }
+        } on FormatException {
+          // Invalid UTF-8 cannot match a sensitive key. Do not let telemetry
+          // sanitization prevent the HTTP client from handling the request.
+        }
+        return part;
+      })
+      .join('&');
+  return url
+      .replace(
+        userInfo: url.userInfo.isEmpty ? null : 'REDACTED:REDACTED',
+        query: url.hasQuery ? query : null,
+      )
+      .toString();
+}
--- a/lib/src/integrations/http_url_redaction.dart
+++ b/lib/src/integrations/http_url_redaction.dart
@@ -1,9 +1,9 @@
-/// Redacts the HTTP span URL without changing the request or Faro event URL.
+/// Redacts HTTP span and event URLs without changing the request URL.
 ///
 /// OTel HTTP client URL redaction rules (the query-key list is Development):
 /// https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span
 /// Preserves the encoding/order of non-sensitive query parameters.
-String redactHttpSpanUrl(Uri url) {
+String redactHttpUrl(Uri url) {
   const sensitiveKeys = {
     'X-Amz-Signature',
     'X-Amz-Credential',
--- a/lib/src/integrations/http_url_redaction.dart
+++ b/lib/src/integrations/http_url_redaction.dart
@@ -1,10 +1,12 @@
-/// Redacts HTTP span and event URLs without changing the request URL.
+/// Redacts HTTP span, event and fallback error-log URLs.
+/// The request URL is unchanged.
 ///
 /// OTel HTTP client URL redaction rules (the query-key list is Development):
 /// https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span
 /// Preserves the encoding/order of non-sensitive query parameters.
 String redactHttpUrl(Uri url) {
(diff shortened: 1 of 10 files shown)


lib/src/integrations/http_url_redaction.dart
```diff
@@ -0,0 +1,38 @@
+/// Redacts HTTP span and event URLs without changing the request URL.
+///
+/// OTel HTTP client URL redaction rules (the query-key list is Development):
+/// https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span
+/// Preserves the encoding/order of non-sensitive query parameters.
+String redactHttpUrl(Uri url) {
+  const sensitiveKeys = {
```
domasx2, line 8: The sensitive-key set for query parameter redaction is hard-coded with no way for applications to extend it, which means bearer tokens, API keys, and passwords in query parameters remain unredacted in `url.full` and `http.url`. This contradicts what users would expect from a credential-redaction feature; consider making the set configurable and including common credential patterns in the defaults.

_Reply posted by Claude on Domas's behalf._

domasx2, line 8: The `sensitiveKeySet` is hard-coded to a fixed set of parameters with no way for an app to extend it. This leaves bearer tokens, API keys, and passwords passed as query parameters stored verbatim in `url.full` and `http.url`, which contradicts the intended credential redaction. Make the set extensible and include common credential parameter names like `token` and `password`.

_Reply posted by Claude on Domas's behalf._

domasx2:
The `network_error` log still includes unredacted `innerContext.uri` in `userAttributes['url']`, allowing credentials and signature query parameters to reach Faro when presigned or credentialed URLs fail—contradicting the redaction policy documented in this PR. Redact the URL in the `network_error` log to align with the documented policy.

_Reply posted by Claude on Domas's behalf._

domasx2:
The `network_error` log still includes unredacted `innerContext.uri` in `userAttributes['url']`, allowing credentials and signature query parameters to reach Faro when presigned or credentialed URLs fail—contradicting the redaction policy documented in this PR. Redact the URL in the `network_error` log to align with the documented policy.

_Reply posted by Claude on Domas's behalf._

domasx2 (dismissed):
claude flagged 2 security issues, but otherwise lgtm

lib/src/integrations/http_url_redaction.dart
robert-northmind: I will add this "credential-redaction feature" as a new config option in a follow-up PR.
To keep this PR focused. but thanks for the feedback :) new PR coming soon.

robert-northmind: added more common param names now. And for the config option, I will handle that in a follow-up PR.

aranhave (approved):
LGTM! 🏅

PR #350, opened 2026-09-11
feat(http): configure sensitive query parameter redaction

robert-northmind:
## Description

Applications can add sensitive query parameter names through `FaroConfig.sensitiveHttpQueryParameters`. For example, `{'customer_code', 'checkout_session'}` redacts those values alongside the existing defaults in HTTP spans, fetch events and fallback network-error logs, while preserving the request sent to the server.

Names are copied into an immutable set and matched exactly after decoding, with case preserved. Omitted, null and empty inputs retain all built-in protections. The policy is installed before initialization awaits and resolved per request, so existing clients see the configured names and each in-flight request uses one consistent redacted URL.

## Related Issue(s)

Related to #346. Follow-up to merged #349; targets `main`.

## Type of Change

- [x] 📝 Documentation

## Checklist

## Additional Notes

Tests were written before implementation for configuration semantics, mutation, matching edge cases, initialization/reset behavior and consistent span/event/fallback-log propagation. Loopback tests verify that the server receives the original URI.

All six pre-release checks passed. Android runs of the example app and Flutter demo verified 10 HTTP span/event pairs and two fallback logs against stored telemetry using dummy query credentials. This setting is limited to automatic HTTP URL fields; arbitrary exception messages, log text, custom attributes and WebView URLs remain outside its scope.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Application-specific HTTP query values can be redacted** alongside the
+  built-in credential parameters. The same policy applies to spans, HTTP
+  events and fallback network-error logs without changing requests. See the
+  Reference docs for configuration and matching behavior
+  ([#346](https://github.com/grafana/faro-flutter-sdk/issues/346)).
+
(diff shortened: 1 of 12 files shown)


aranhave (approved):
LGTM

